PR #48, opened 2018-03-28
[Add] EXII Classes, Jeanne (Themed)

Groovman:
I added the new EXII classes as available classes and the Jeanne (Themed) character. I also added a 1000% DA buff to Seox and Albert so they don't benefit from additional DA weapon effects, and all Jeanne versions have their crit support skill added. The translation for characters is also updated. character generator also works on windows now and from the director it is in.


hsimyu:
I confirmed your updates. Thanks!

PR #50, opened 2018-03-31
[Add] Robomi Z characters and weapons, update Magna2 weapons

Groovman:
Hi, I added the new characters and weapons from Robomi Z. I also updated the shiva trident and removed weapon skills for lvl 120 from non MLB MagnaII weapons as it already is with the Metatron bow.


hsimyu:
The commit for RobomiZ is no problem, thanks!
The second commit, for "magnaSanteS" (Magna Trium (S)), is inappropriate because Trium skill enhances both Double Attack and Triple Attack.
The new skill for the shiva weapon (破壊) only enhances Triple Attack.
"magnaLesserSanteS" will be appropriate but this have not been implemented yet.
I'm planning to add it and other new skills next week.

Groovman:
Ok give me a few minutes to fix it

Groovman:
Should now work as it should.

hsimyu:
Okay, thanks!
I'll test it.

PR #52, opened 2018-04-08
Detective Conan, Dual weapon prof, Wyrmsword type

Groovman:
Hi, I added the Detective Conan character and weapons. I also fixed the pattern matching for dual weapon profession characters. The issue about the Wyrmsword element type that is wind instead of earth is also fixed and can be closed.


hsimyu:
Thanks!!!!!!! (sorry for my slow response)

PR #58, opened 2018-04-17
Alexiel, SR characters, 100 Moon weapons, small fixes

Groovman:
Hi, I added Alexiel and her gatcha weapon and also the missing SR characters. I also added the 100 moon FLB weapons. I fixed the small issues with the non-FLB cerberus order that was reported. I added the missing character pictures to the image url list that was also reported.


hsimyu:
I think there are no problems except for e156b72.
For this commit, we have to confirm the amount of the two new skills.
Probably, primacy (L) is ok, it will be the same as normal might + dual edge.

Do you have any information about 刃界(Bladeworld)?
It's completely new skill, so I want to confirm the estimation of the amount of it before implementation.

Groovman:
No, sorry found no info on it not the wiki, twitter or gamewith, probably too early to get concrete info on it. I assumed the unique name just comes from the unique combination.

Groovman:
Primacy is confirmed here
https://xn--bck3aza1a2if6kra4ee0hf.gamewith.jp/article/show/99661
It seems it is Primacy (LL) with ATKup SL15 = 20%  and with DA SL10 = 7.0% which would be a new rank for normalNite, I looked at the numbers and I think it the lower start is 2 times the difference between the previous ranks, and the steps are 0.6, 2 times the step increase from L to M (If I take a lower starting point the step size would be far too large). With a fix point at SL10 of 7.0 I get something like this: 
`"normalNite": {
        "S": [0.4, 0.6, 0.8, 1.0, 1.2, 1.4, 1.6, 1.8, 2.0, 2.2, 2.4, 2.6, 2.8, 3.0, 3.2],
        "M": [0.7, 1.0, 1.3, 1.6, 1.9, 2.2, 2.5, 2.8, 3.1, 3.4, 3.7, 4.0, 4.3, 4.6, 4.9],
        "L": [1.0, 1.4, 1.8, 2.2, 2.6, 3.0, 3.4, 3.8, 4.2, 4.6, 5.0, 5.4, 5.8, 6.2, 7.0],
        "LL": [1.6, 2.2, 2.8, 3.4, 4.0, 4.6, 5.2, 5.8, 6.4, 7.0, 7.6, 8.2, 8.8, 9.4, 10.0],
    },`
The 2nd skill still has no info.

hsimyu:
OK, I merged it and I'll announce on twitter that the 2nd skill (Bladeworld) is being unconfirmed at present.

PR #85, opened 2018-06-15
Add Production Optimized Dockerfile

FabulousCupcake:
#### Description
- Updates `package.json` and `package-lock.json`
  - Move packages required for production build to `dependencies`.
  - Remove unused packages (`file-loader`, `install`, and `npm`).

- Updates `Dockerfile` and `docker-compose.yml`
  - Utilize [multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/) for a leaner final image.
  - Use alpine based builder image to speed up build time.
  - Use node v10 for the building stage so we can utilize [`npm ci`](https://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable) to speed up the build process.
  - Use `nginx` as the final base image, since the build results in a static webpage.

#### Notes
- Create a separate Dockerfile that runs the [image download scripts](https://github.com/MotocalDevelopers/motocal/tree/master/scripts) during the image building process?


glassmonkey:
@FabulousCupcake 
Thankyou for update the Dockerfile. 

Could you can change node 8.x.x  and leave item written node in docker-compose.yml

because our production environment is 8.x.x and and Dockerfile and docker-compose.yml assist for development, so  can't remove `npm run start`.

hrhyt:
@FabulousCupcake

Thank you for your pull request.

Do you mean you want to build with docker?

Can you change the name of `docker-compose.yml` and `Dockerfile`?

build can be executed with the following command.

```
docker-compose.exe -f docker-compose_xxx.yml up
```

> Create a separate Dockerfile that runs the image download scripts during the image building process?

It is very useful to download images in the build process.

FabulousCupcake:
@glassmonkey @hrhyt 

Thank you for the feedback.

It seems the current `Dockerfile` and `docker-compose.yml` is used for development process. I did not realize that.

I will close this PR and create another one that will not touch the existing `Dockerfile` and `docker-compose.yml`. Instead, they will in a separate file and be used only for deployment (in this case Heroku).

PR #96, opened 2018-06-19
Added weapons and character for Patch 2018-06-18

Groovman:
Same as every patch :) Wulf has 100% DA/TA to simulate his guaranteed triple attack support skill.


sys-cat:
@Groovman thanx patch !! merge and closed.

PR #115, opened 2018-07-15
Recent Patches + TA Update

Groovman:
Hi there, 

here are the recent character and weapons updates. Every added uncap from the primal update is still not fully tested, but I didn't want to delay the pull request more, because I had no time for testing everything. I also updated the TA numbers as they should be going by the formula.


hrhyt:
全てのテストはされていないとのことでしたので、Productionブランチに反映する前に更新履歴に書いた方が良いかもしれませんね。

PR #120, opened 2018-07-25
Latest patch, Character Reviews, Destruction Skill update

Groovman:
Hi there,

Latest Character + Weapons for Patch 2018-07-18, also updated the translation for Zooey (Summer) to Zooey (Grand). 

I reviewed some characters (5* and limited, still need to review all normal 4*) for data that is outdated currently and updated Anila's weapon prof in the character data so it gets extracted correctly and her support ability from 5% to 10% so her 5* is now accurate. Mikasa's weapon prof was wrongly formated in the character data, is now fixed so the script extracts it correctly. Lancelot 5* support skill DA also updated from 5% to 10%. Updated the character script to patch Yuel and Yuel(Event) DA values because they get own DA from their support ability. Lancelot shouldn't need values patched because his support skills buffs all including himself not only his allies.

The Destruction was 2.0 the wiki and the blog linked in the commit as source have it as 2.5 so updated all values to 2.5 from 2.0.


Groovman:
Added Character + Weapons for Patch 2018-07-31, also reordered characters and added 5* art to characters with a 5* uncap.

PR #127, opened 2018-08-04
Recent patches

Groovman:
Added the characters for Patch 2018-08-04, cleaned up the arm script, and granni bow was missing the new sensei skill.


Groovman:
Added the recent patches, was waiting for data on Draco Horn for 2 days. I estemated the atk/hp for the 4*, average for a spear from 3* to 4* is +45HP and +450ATK, spear is more on offense so added +42HP and 470ATK.

Groovman:
Added Scathacha and Gram uncaps.

Groovman:
Added the necessary patching for the updated DATA values from #129

PR #132, opened 2018-12-21
Akasha weapons, new characters, new job, support abilities, updated scripts, minor fixes

Groovman:
Hi there,

I added the new Akasha weapons with the skill values that are on the wiki and the new characters from patch 2018-12-17. Added SSR characters support abilities where it was possible. Updated the bash scripts to python scripts and added the possibility to pull the images directly from the game. Image erors have now a No Image alternative instead of an error image. Pulling the images from the game might be a bannable offense so don't use it if you don't have to, Updated the readme because of that and updated/translated some parts. MIT license has no liability anyway. Moved the url lists to the txt_source folder. Did some general code cleanup so it a little bit better readable and extendable.

OrkunKocyigit added ougi buff to summons tab.

Happy holidays everyone!


kei-gbf:
Hello,
Glad to see active developer, and thanks a lot updates!

I am new to see this repo. I also had similar works in local then I can review a bit.
It's large commits for me to follow all, so this is a part of I've noticed.


Akasha weapon must be only one. this does not mean "no effect".
in the game User Interface, it's an impossible to set two Akasha weapons in the same grid.
When players try to set other akasha weapon, the game ask players to replace it.

so the boolean flag should be like this?

```javascript:global_logic.js
var isAkashaIncluded = false;

 /* snip */

  else if (stype == 'akasha' && !isAkashaIncluded) {
    // オメガウェポン   <--- XXX: remove the line comment, this means "Omega weapon" in Japanese
    isAkashaIncluded = true;
```

kei-gbf:
Hello,
Glad to see active developer, and thanks a lot updates!

I am new to see this repo. I also had similar works in local then I can review a bit.
It's large commits for me to follow all, so this is a part of I've noticed.

```javascript:global_logic.js

  else if (stype == 'akasha') {
    // オメガウェポン   <--- XXX: remove the line comment, this means "Omega weapon" in Japanese
```

EDIT:
Akasha weapon must be only one. this does not mean "no effect".
in the game User Interface, it's an impossible to set two Akasha weapons in the same grid.
When players try to set other akasha weapon, the game ask players to replace it.

The akasha isIncluded flags is for limit one weapon per weapons types.
How motocal make it exclusive? or duplicate the effects?

Groovman:
Hi,

Yeah I saw your commit history with the single flag for the weapons. Mine was still just a modified copy of the ultima weapons at the time. But it did not solve the problem that 2 or more akasha weapons are getting recommended. I just took a look at it again and basically you can put the flag before the loop over the party and as long as enough combinations are there only combinations with 1 akasha weapon are getting recommended. You can not exclude weapons on basis that another weapon is present, you would need to change the code that generates all the combinations, but that is far more work.

Comment obviously gets removed :)

kei-gbf:
> Yeah I saw your commit history with the single flag for the weapons. Mine was still just a modified copy of the ultima weapons at the time. But it did not solve the problem that 2 or more akasha weapons are getting recommended. I just took a look at it again and basically you can put the flag before the loop over the party and as long as enough combinations are there only combinations with 1 akasha weapon are getting recommended. You can not exclude weapons on basis that another weapon is present, you would need to change the code that generates all the combinations, but that is far more work.

After I tested the single flag code, I had noticed that too.

Two akasha weapons is a rare case if users knows the weapon's specification,
they won't try two or more akasha weapons, that is a kind of abuse.
so this can be low priority(or another issue). I just note this for share.

There are use-cases exclusive/inclusive 
- 1. If user wants to see which akasha weapon is better for their characters. exclude for the case.
  (Akasha weapon supports two fav types, and character also has two)
- 2. User may want to see IF two or more akasha weapons is allowed. 
  (inclusive for the assumed situation)

The first, I've expected was the first use case, but I understand that single flag did not solve it.
I did not see the combinations. Thank you for explaining those info.

I think the current included flag is ok enough, this is not so critical.

> Comment obviously gets removed :)

Sorry, the post was my mistake.

Groovman:
Motocal generates all combinations of 10 weapons it can do, if two or more Akasha weapons are selected, that includes some with only one weapon and all with multiple. If you have a first match for a favorite weapon the akasha weapon gets included in the calculation, following akasha weapons are skipped resulting in a much worse damage value  Basically a grid with 2 akasha weapons is technically a grid with 9 weapons and will be worse, the best akasha weapon will come out top.

So the weapon that is better than another for the given team comp will be higher in the list. Main problem currently is that everything is stupidly in favor of the axe because any weapon has the 20% EX modifier, but the axe has 10% TA which is great in the formula compared to some modifiers like normal Crit, 10% extra EX mod, 10% extra HP and debuff success, last 2 modifiers don't even affect the damage at all.

People need to understand what they throw into motocal and what the values mean they get out of it. But that is a problem with any theorycrafting tool.

kei-gbf:
>  2 akasha weapons is technically a grid with 9 weapons and will be worse

I see, ok. This is a test case for the abuse. Just skip for now, it's a rare as real use case.

- in case two akasha weapons and other weapon x8
  - two akasha weapons is the top, second akasha weapon skill is skipped but included arm attack/hp.
- in case two akasha weapons and other weapon x9
  - Yes, akasha axe came on the top as you told.

I saw where you said "that is far more work." it's in the calcCombinations function.
there were isCosmosArray and so on.

> ... But that is a problem with any theorycrafting tool.

I agree. Users can avoid two akasha weapons by their usage.

scripts/arm_data_converter.py
```diff
@@ -1,8 +1,6 @@
-# -*- coding:utf-8 -*-
-import json
-import re
-import csv
+import os, json, csv, re, inspect
```
kei-gbf: Do you care coding style? if yes ...

In Python coding style guide (known as PEP8 or any others)
`Imports should usually be on separate lines:`
One of reasons, when developers read the DIFF, separate lines import is better.

This may be mixed another script languages cultures.
It's ok little thing since this project does not have Contributing guide lines.

kei-gbf:
ah, sorry i noticed I did not have the write access for reviewers ... (to pass the merge)

hsimyu:
Hi, the original author is here.
Thank you for your contribution and greatly improvement of update workflow.

Actually, there are no active maintainers on this repository except for you.
So, I think this PR should be okay to be merged.

hsimyu:
Btw, could you join the 'MotocalDevelopers' organization? @Groovman and @kei-gbf 
If you joined, I can add permission to press 'Approve'.

scripts/arm_data_converter.py
Groovman: I work mostly in Webstorm for motocal I didn't run the python scripts with auto format and auto optimize with Pycharm, but with Webstorm and that doesn't do formating correctly for whatever reason.

Groovman:
@hsimyu can you resend me the invitation? github says it is revoked.

scripts/arm_data_converter.py
kei-gbf: Any Python IDE/editors can warn unused module imports. (or maybe extra option for Linter)
I saw now 7e7c2b8 some unused imports are removed. 👍

kei-gbf:
> Btw, could you join the 'MotocalDevelopers' organization? @Groovman and @kei-gbf
> If you joined, I can add permission to press 'Approve'.

Nice to see you here, I heard the news the development stopped
but I did not know where to contact. I just know github and PR develop cycle.

Can you send invite me? (there was no join buttons on my side)

hsimyu:
@Groovman @kei-gbf 
I sent invitations now.

src/global_logic.js
```diff
@@ -1955,7 +1955,20 @@ module.exports.treatSupportAbility = function (totals, chara) {
                         totals[key]["TABuff"] += 0.05;
                     }
                     continue;
-                case "taiyou_sinkou":
+                case "daBuff_fist":
+                    if (totals[key].isConsideredInAverage) {
+                        // ドラフと種族不明のみキャラ攻刃
```
kei-gbf: This is also a wrong comment.

kei-gbf:
@hsimyu 
> I sent invitations now.

I joined, thank you!

kei-gbf:
btw, I can not review those all commits, or take a long time. it's too large for me.
as I saw, the most of I noticed are all little things.

so if it's ok, please merge. people may be waiting motocal update.

Groovman:
Everything cleaned up and added the new DATA values for earth characters from issue #129.

kei-gbf:
Question about akasha bow skill value. I did not test myself though,
I found another data it say: value is "about" 20% attack ratio is 50% (Ougi effect is about 60%/ratio 30%)
I can't find which commit had the code. so I comment here.
Is it now 10% and ratio 45% ?

[虚ろなる神器]虚空の歪弦
http://gbf-verification.blog.jp/archives/14717421.html

kei-gbf:
> Question about akasha bow skill value. .. snip..

I've found now the source, from English wiki and the verification was posted in twitter.
10% / 45% Critical Hit Rate Up.

kei-gbf:
I suggest postpone it until additional test published.
I think the current skill values are ok for now.

reasons
- nothing special motivation to change it in this PR
  - another values maybe wrong, or at least we don't know.
- 10%/45% is a minimum value in the sets
  - the evolution of the bow will not down

It is not so critical problem should be fixed before release.
merge first, and let people try then get feedback.


about the skill's different data.
it's a lack of the information how they had tested.
- about 10-20% chances
  - the numbers of trial
  - error
  - possibility of missing another bonus 
- 45% or 50% damage
  - include the additional 30000 damages ?
  - damage cap??
  - etc ...

If there was a similar case in the past, or guide line for additional test, we can follow the workflow.
for now, this is just my suggestion.

I don't have the permission to merge, let wait @hsimyu can merge it.

kitsuki00:
src/global_logic.js
```
458     - var expectedAttack = 3.0 * taRate + (1.0 - taRate) * (2.0 * daRate + (1.0 - daRate))
---
501      +var expectedAttack = 3.0 * taRate + (1.0 - taRate);
502     +expectedAttack *= 2.0 * daRate + (1.0 - daRate);
```
calculation result here has changed.
sorry... I will leave it because I do not know the correct calculation.

src/global_logic.js
```diff
@@ -1941,6 +1970,13 @@ module.exports.treatSupportAbility = function (totals, chara) {
                     // ザルハメリナのHPを参照する
                     totals[key]["charaHaisui"] += module.exports.calcHaisuiValue("charaHaisui", "L", 20, totals[key]["remainHP"]);
                     continue;
+                case "envoy_meditation":
+                    var elements = module.exports.checkNumberOfElements(totals);
+                    // 4種族なら50%, それ以外なら種族数*10%
```
kei-gbf: Comment translation, 種族(race) 属性(element)  and the number is not 50%/10% here.

`// 属性数 * 攻撃15% DA10% TA3%` or just can be removed.

Groovman: 👍

src/global_logic.js
```diff
@@ -741,6 +741,35 @@ module.exports.checkNumberofRaces = function (chara) {
...
+    var includedElements = {
+        "fire": false,
+        "water": false,
+        "earth": false,
+        "wind": false,
+        "light": false,
+        "dark": false,
+    };
+
+    // indの初期値も1からで良い
```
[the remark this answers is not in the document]
Groovman: 👍

src/global_logic.js
```diff
@@ -1975,7 +1970,7 @@ module.exports.treatSupportAbility = function (totals, chara) {
                     continue;
                 case "envoy_meditation":
                     var elements = module.exports.checkNumberOfElements(totals);
```
kei-gbf: Now (ind < 4) check was removed. the function may return 5 and 6? (for sub members)

I do not know yet about the "totals" data structure.
I worry is if a party like [dark, dark, dark, dark, water, fire] return 3
but front members elements are only dark.


```javascript
var elements = Math.max(4, module.exports.checkNumberOfElements(totals));
```

then code below does not need to check element >= 4 every times.
```javascript

totals[key]["normalBuff"] += (elements * 0.15);
totals[key]["DABuff"] += (elements * 0.10);
totals[key]["TABuff"] += (elements * 0.03);
```
for more clean up (hand write code here in comment form, not tested)

kei-gbf:
```javascript
var elements = Math.min(4, module.exports.checkNumberOfElements(totals));
```

then code below does not need to check elements >= 4 every times.
```javascript

totals[key]["normalBuff"] += (elements * 0.15);
totals[key]["DABuff"] += (elements * 0.10);
totals[key]["TABuff"] += (elements * 0.03);
```
for more clean up (hand write code here in comment form, not tested)

src/global_logic.js
```diff
@@ -741,6 +741,35 @@ module.exports.checkNumberofRaces = function (chara) {
     return races
 };
 
+module.exports.checkNumberOfElements = function (totals) {
+    // check num of races
```
kei-gbf: // check num of "elements" here.

src/global_logic.js
Groovman: 👍

src/global_logic.js
```diff
@@ -750,14 +750,9 @@ module.exports.checkNumberOfElements = function (totals) {
         "dark": false,
     };
 
-    // indの初期値も1からで良い
-    var ind = 0;
     for (var key in totals) {
         if (totals[key].name != "" && totals[key].isConsideredInAverage) {
-            if (ind < 4) {
```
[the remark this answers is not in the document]
Groovman: Yeah remembered also why I left the index var there in the first place :D

src/global_logic.js
Groovman: I am going to translate all comments. give me a few minutes and I am going to sanitize the code a little bit further.

src/global_logic.js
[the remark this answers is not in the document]
kei-gbf: If "totals" var contains sub members, it's an array?
one more things I worry there, it is the order of elements in the for-in loop.
(I found for-in loop are used a lot in the file)

for-in is for object properties. using for-in to array is not safe.
Although, the code may work correctly now.
If the order is an important, maybe  for-of   or  .forEach is for Array.

`Looping over objects with a for in loop will include properties that are inherited through the prototype chain. This behavior can lead to unexpected items in your for loop.`
https://eslint.org/docs/rules/guard-for-in

If current code worked then it's ok, there are a lot for-in loops and working.
Another PR in future can solve it. (this PR has many topics already)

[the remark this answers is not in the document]
kei-gbf: Ah, the quote link was not about the order of for-in. but that is also another issue.
about the order of properties in for-in loop I could find info:

https://johnresig.com/blog/javascript-in-chrome/
`The mechanics of enumerating the properties … is implementation dependent`

src/chart.js
```diff
@@ -106,24 +106,24 @@ var HPChart = CreateClass({
...
-        for (var i = 1; i < data.length; i++) {
-            csvData.push(data[i]);
+        // where d can not be null but conservatively returns
+        if (d != null) {
+            var data = d[this.state.sortKey];
+            var csvData = [];
+            var header = [];
+            header.push(data[0][0]);
+            for (var i = 1; i < data[0].length; i++) {
+                // 不要な改行が見出しに含まれておりCSVが崩れるので消す
```
kei-gbf: // Remove unnecessary line feed letters which break CSV format.

src/global_const.js
```diff
@@ -1264,36 +1263,41 @@ var skillAmounts = {
...
-    // 楚歌 効果が1本のみとなる為、通常・マグナ攻刃とは別途定義
+    // Since it is only one Soka effect, it is usually defined separately from Magna attacking blade
     "normalSoka": {
         "M": [3.0, 3.0, 4.0, 4.0, 5.0, 5.0, 6.0, 6.0, 7.0, 8.0, 8.0, 8.0, 8.0, 8.0, 8.0, 8.0, 8.0, 8.0, 8.0, 8.0],
     },
     "magnaSoka": {
         "M": [3.0, 3.0, 4.0, 4.0, 5.0, 5.0, 6.0, 6.0, 7.0, 8.0, 8.0, 8.0, 8.0, 8.0, 8.0, 8.0, 8.0, 8.0, 8.0, 8.0],
     },
-    // 先制スキルは加護が乗らない　Slv11~14までは仮入力
+    // Pre-emptive skills
```
kei-gbf: Important info in this comment was:
- Slv 11..14 are temporary values.

means those values are not verified, it was outdated comment ? or leave it.
looks "sensei" has values in slv 11..14 if they are verified then ok. can remove the tail of comment.

src/global_logic.js
```diff
@@ -147,19 +147,19 @@ module.exports.makeSummonHeaderString = function (summon, locale) {
...
+    // Damage calculation
+    var def = enemyDefense != undefined ? enemyDefense : 10.0;
     var damage = Math.ceil(Math.ceil(summedAttack / def) * totalSkillCoeff) * criticalRatio;
     var overedDamage = 0;
 
     var limitValues = [[600000, 0.01], [500000, 0.05], [400000, 0.60], [300000, 0.80]];
 
     for (var index = 0; index < 4; index++) {
-        // 減衰ライン算出
+        // Damping line calculation
```
kei-gbf: as direct translation it's ok. but is that ok as English? it can be "border"
I am a native Japanese, and we use "line" on this context as a short word of "borderline".
Japanglish? (Japanese use English word in a different way)

I found English wiki use "Damage cap" for this term
https://gbf.wiki/Damage_cap

src/global_logic.js
```diff
@@ -267,25 +268,25 @@ module.exports.calcCriticalArray = function (normalCritical, _magnaCritical, nor
         damageRatio.push(0.5);
     }
 
-    // 通常技巧配列は[確率1, 確率2, 確率3, ... ]という形式で渡される
+    // Normal skill array is passed in the form of [probability 1, probability 2, probability 3, ...]
```
kei-gbf: 技巧(Verity)  so ... Normal-verity array ?
skill is スキル

src/global_logic.js
```diff
@@ -528,66 +532,66 @@ module.exports.calcBasedOneSummon = function (summonind, prof, buff, totals) {
...
+        ougiDamageLimitByNormal = ougiDamageLimitByNormal <= 0.30 ? ougiDamageLimitByNormal : 0.30;
+        var ougiDamageLimitByMagna = totals[key]["magnaOugiDamageLimit"] * totalSummon["magna"];
+        ougiDamageLimitByMagna = ougiDamageLimitByMagna <= 0.30 ? ougiDamageLimitByMagna : 0.30;
         var ougiDamageLimit = buff["ougiDamageLimit"] + totals[key]["ougiDamageLimitBuff"];
         ougiDamageLimit += ougiDamageLimitByMagna;
         ougiDamageLimit += ougiDamageLimitByNormal;
         ougiDamageLimit += ougiDamageLimitByExceed;
 
-        // damageは追加ダメージなしの単攻撃ダメージ(減衰・技巧補正あり)
+        // damage is single damage without additional damage (with attenuation and skill correction)
```
kei-gbf: 単攻撃 means SA (single attack), like DA, TA
"damage" is a single attack damage without ...

src/global_logic.js
```diff
@@ -528,66 +532,66 @@ module.exports.calcBasedOneSummon = function (summonind, prof, buff, totals) {
...
         var damage = module.exports.calcDamage(summedAttack, totalSkillCoeff, criticalRatio, prof.enemyDefense, additionalDamage, damageUP, damageLimit);
 
-        // クリティカル無しの場合のダメージを技巧期待値の補正に使う
+        // Use damage in case of no critical to correct skill expectation
         var damageWithoutCritical = module.exports.calcDamage(summedAttack, totalSkillCoeff, 1.0, prof.enemyDefense, additionalDamage, damageUP, damageLimit);
 
-        // 実質の技巧期待値
+        // Real skill expected value
```
kei-gbf: 技巧(Verity) here
confused comment

スキル ... skill  This is ok
技巧 ... translate this to skill is No. it's one of skill names.

src/global_logic.js
```diff
@@ -720,11 +727,12 @@ module.exports.checkNumberOfRaces = function (chara) {
         "seisho": false,
         "unknown": true,
     };
-    // ジータがいるのでunknown枠は常にtrue
-    // indの初期値も1からで良い
+    // Since there is a Djeeta, the unknown frame is always true
```
kei-gbf: 枠 "frame" direct translation is ok but
Here on the context of JavaScript
it's object's "property" or "field"

src/global_logic.js
```diff
@@ -720,11 +727,12 @@ module.exports.checkNumberOfRaces = function (chara) {
...
+    // The initial value of ind can also start from 1
     var ind = 1;
     for (var key in chara) {
         if (chara[key].name != "" && chara[key].isConsideredInAverage) {
+            // only front line
```

src/global_logic.js
```diff
@@ -958,12 +950,12 @@ module.exports.addSkilldataToTotals = function (totals, comb, arml, buff) {
...
     var index = 0;
     for (var key in totals) {
-        index = (index + 1) | 0;
+        index = index + 1 | 0;
         var isBahaAtIncluded = false;
         var isBahaAthpIncluded = false;
         var isBahaHpIncluded = false;
 
-        // オメガウェポンは別スキルなら複数効果発動可能
+        // Omega Weapon can activate multiple effects if it is different skill
```
kei-gbf: translate "Omega weapon" is ok for English ?
I note this
Japanese: Magna, Omega are 
English: Omega, Ultima (https://gbf.wiki/Category:Omega_Weapons  and  https://gbf.wiki/Ultima_Weapons)
in Motocal source code, magna and omega are used.

src/global_logic.js
```diff
@@ -977,441 +969,444 @@ module.exports.addSkilldataToTotals = function (totals, comb, arml, buff) {
...
-                                totals[key]["race"] === "seisho") {
-                                totals[key]["bahaAT"] += comb[i] * skillAmounts["bahaAT"][amount][slv - 1];
-                                isBahaAtIncluded = true;
-                            } else {
-                                var bahatype = skillname.split("-");
-                                if (bahamutRelation[bahatype[1]]["type1"] == totals[key]["race"]) {
+                        // Baha, cosmos and omega weapons have no attribute relation
+                        if (stype == 'bahaAT') {
+                            if (!isBahaAtIncluded) {
+                                // バハ短剣など
```
kei-gbf: // Baha dagger etc.

src/global_logic.js
```diff
@@ -1995,7 +1991,7 @@ module.exports.treatSupportAbility = function (totals, chara) {
                     break;
             }
 
-            // 技巧系処理
+            // Skill processing
```
kei-gbf: this "Skill" is also "Verity" here.
how about Critical processing ?

Motocal use criticalBuff field as 技巧(Verity)

src/chart.js
Groovman: missed that one

src/global_const.js
Groovman: Yeah already confirmed, I check the skill values from time to time if something changed and Pre-emptive was confirmed some time ago, critical or something was the last thing I updated with updated wiki values when I remembered right.

src/global_logic.js
Groovman: Normal critical skill array, normal crit is handled via an array with each instance that can crit while, magna crit is a single value that stacks with every instance in the grid.

src/global_logic.js
```diff
@@ -464,55 +465,58 @@ module.exports.calcBasedOneSummon = function (summonind, prof, buff, totals) {
...
+        var expectedAttack = 3.0 * taRate + (1.0 - taRate) * (2.0 * daRate + (1.0 - daRate));
 
         if (totals[key]["typeBonus"] == 1.5) {
-            // 与ダメージ上昇サポアビは天司スキルと重複しない（効果の高い方のみ発揮）
-            var damageUP = (totals[key]["tenshiDamageUP"] > totals[key]["charaDamageUP"]) ? totals[key]["tenshiDamageUP"] : totals[key]["charaDamageUP"];
+            // Supplemental damage rise support ability does not overlap with Tenshi skill (the strongest effect overwrites the lesser)
+            var damageUP = totals[key]["tenshiDamageUP"] > totals[key]["charaDamageUP"] ? totals[key]["tenshiDamageUP"] : totals[key]["charaDamageUP"];
 
-            // 通常別枠の技巧スキル配列を生成
+            // Generate a special skill skill array normally
```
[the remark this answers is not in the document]
Groovman: "Generate normal critical skill arrays." would make the most sense then

src/global_logic.js
Groovman: 👍

src/global_logic.js
Groovman: Yeah should probably be "expected critical skill ratio" Verity being critical skill but yeah google translate doesn't know that.

src/global_logic.js
Groovman: 👍

src/global_logic.js
Groovman: I like omega here more because if I search parts in the logic I mostly search via the skill prefix in the skill list and that is baha, cosmos and omega. So it is just easier to find those parts when they have similar names.

src/global_logic.js
Groovman: 👍

src/global_logic.js
Groovman: 👍

src/global_const.js
kei-gbf: :+1: ok, I did not know the value details, so just a confirmation.

src/global_logic.js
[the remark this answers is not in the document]
kei-gbf: Yes, critical skill is better for this translation.
"Verity" I found was in game (when I change to English version)

public/index.html
```diff
@@ -47,7 +47,7 @@
         <% if (htmlWebpackPlugin.options.ADSENSE_AD_CLIENT && htmlWebpackPlugin.options.ADSENSE_AD_SLOT_PC2) { %>
             <div id="adsense-original-div-pc-2" class="hidden">
                 <script async src="//pagead2.googlesyndication.com/pagead/js/adsbygoogle.js"></script>
-                <!-- 元カレ計算機 PC版上部 -->
+                <!-- motocal PC version upper part -->
```
kei-gbf: I wonder why two different translations for 上部 (I got google translate shown)
"upper part" and "top part", more intention words are "page top" or "header" (Web terms)
or PC/smartphone version is an enough.

It's a short html file, Readers knows it's on the top part without the comment.
An important info is switch code for PC or Smartphone.

Groovman: Yeah header... silly me :D

kei-gbf:
Review for f8c14ea

- [x] :+1: errors showing the "Sorry" sticker
- [x] Check characters 'sex' field.
  - characters data table in wiki: http://gbf-wiki.com/index.php?%A5%C6%A1%BC%A5%D6%A5%EB%2F%C1%B4%A5%AD%A5%E3%A5%E9%A5%AF%A5%BF%A1%BC%B0%EC%CD%F7
  - check script:  https://gist.github.com/kei-gbf/5ad8fbb196c0d8e8a15fcadb43d85c5c
    - This script checks charaData.json 'sex' field match to wiki's character table.
- [ ] four python scripts
  - [x] :+1:  operate independent of the path they are called from
  - [ ]  Cons: four copy-and-pasted code increases the maintenance cost x4 times
    e.g. If one had a bug, then have to fix all copied codes.
    the difference can be parameter? I can show a sample
    - https://gist.github.com/kei-gbf/8e878356e3ddee7476d026ec437343eb

----
## Characters data reports

output of the check script above.

```
NG    1 NO KEY FOUND M ユーステス(闇属性ver)
NG    2 NO KEY FOUND F フォリア(リミテッドver)
NG    3 SEX MISMATCH F サンダルフォン
NG    4 NO KEY FOUND M ドランク
NG    5 NO KEY FOUND F スツルム
NG    6 NO KEY FOUND F ジャンヌダルク(闇)
NG    7 NO KEY FOUND O ベス(攻撃)
NG    8 NO KEY FOUND O ベス(防御)
NG    9 NO KEY FOUND O ベス(バランス)
NG   10 NO KEY FOUND M ルシウス(火属性)
NG   11 NO KEY FOUND F スーテラ(イベントver)
NG   12 NO KEY FOUND F シャルロッテ(イベントバージョン)
NG   13 NO KEY FOUND M ジャミル(闇属性ver)
NG   14 NO KEY FOUND M ウェルダー(イベントver)
NG   15 NO KEY FOUND F ゾーイ(イベントver)

Total number of table rows: 442 (wiki)
Total number of characters: 453 (charaData.json)
```

Those "NO KEY FOUND" lines are all ok,
key(character name) in wiki table contains additional info like '(リミッテッドver)' etc.
that's why script reported so. it's enough, I could checked the rest data manually.

MISMATCH line is true NG case, it was only one in the json data.

- [ ] Sandalphon is 'female' ?

src/global_const.js
```diff
@@ -1400,15 +1418,15 @@ var supportAbilities = {
...
-        "type": "taiyou_sinkou",
+        "type": "emnity_all_SL10",
         "range": "all",
         "value": 0.00
     },
-    "la_pucelle": {
-        "name": "に背水効果(ジャンヌダルク(闇))",
-        "type": "la_pucelle",
+    "emnity_own_SL20": {
+        "name": "に背水効果(ジャンヌダルク(闇), 黒騎士)",
```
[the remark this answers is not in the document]
Groovman: 👍

Groovman:
> Review for [f8c14ea](https://github.com/MotocalDevelopers/motocal/commit/f8c14ea56a9be4f6e84d9f91ef862e064d2f4179)
> 
> MISMATCH line is true NG case, it was only one in the json data.
> 
> * [ ]  Sandalphon is 'female' ?

Nah that is me making an Oopsie, when I edited the file by hand. :D

Groovman:
> Review for [f8c14ea](https://github.com/MotocalDevelopers/motocal/commit/f8c14ea56a9be4f6e84d9f91ef862e064d2f4179)
> 
> * [x]  Check characters 'sex' field.
>   
>   * characters data table in wiki: http://gbf-wiki.com/index.php?%A5%C6%A1%BC%A5%D6%A5%EB%2F%C1%B4%A5%AD%A5%E3%A5%E9%A5%AF%A5%BF%A1%BC%B0%EC%CD%F7
>   * check script:  https://gist.github.com/kei-gbf/5ad8fbb196c0d8e8a15fcadb43d85c5c
>     
>     * This script checks charaData.json 'sex' field match to wiki's character table.
> * [ ]  four python scripts
>   
>   * [x]  👍  operate independent of the path they are called from
>   * [ ]   Cons: four copy-and-pasted code increases the maintenance cost x4 times
>     e.g. If one had a bug, then have to fix all copied codes.
>     the difference can be parameter? I can show a sample
>     
>     * https://gist.github.com/kei-gbf/8e878356e3ddee7476d026ec437343eb
> 

I like the idea of parameterized scripts, but you could simplify the scripts a little bit because you do not have to add the paths as parameter in my opinion, the scripts are mostly run from the scripts folder. Image script would just need arms/chara and wiki/game as choices, folder structure should not change.

Groovman:
Normal weapon character update, I changed the check_sex script a bit but had no time to look at the new image download script been busy with other stuff, will look at it when I have some more free time and then also update the readme.

kei-gbf:
thanks new update patch!

Akasha weapon exclusion
!isAkashaIncludedGlobal  and   isAkashaIncludedGlobal  is a double checking?
(I do not know the details of logic yet) 

```javascript:
  if (!isAkashaIncludedGlobal || (isAkashaIncludedGlobal && isAkashaIncludedLocal[akashaType])) {
```

| global | local |  expr  | result |  |
|----------------------------|-------------------------|--|--|--|
| false | false | `(!false \|\| (false && false))` | true | if not included yet ... ok |
| false | true | `(!false \|\| (false && true))` | true | never check includedLocal ... ok |
| true | false | `(!true \|\| (true && false))` | false | included but other weapon type | 
| true | true | `(!true \|\| (true && true))` | true | included already ? |

I do not know how motocal should calc this. at least `(isAkashaIncludedGlobal &&` looks redundant.
It's not effect to the result by omit the double checking.


btw, I got an error saving file on scripts/download_charaimages_from_wiki.py 
the split separator is "=" ?

Groovman:
> thanks new update patch!
> 
> Akasha weapon exclusion
> !isAkashaIncludedGlobal and isAkashaIncludedGlobal is a double checking?
> (I do not know the details of logic yet)
> 
> ```
>   if (!isAkashaIncludedGlobal || (isAkashaIncludedGlobal && isAkashaIncludedLocal[akashaType])) {
> ```
> global	local	expr	result	
> false	false	`(!false || (false && false))`	true	if not included yet ... ok
> false	true	`(!false || (false && true))`	true	never check includedLocal ... ok
> true	false	`(!true || (true && false))`	false	included but other weapon type
> true	true	`(!true || (true && true))`	true	included already ?
> I do not know how motocal should calc this. at least `(isAkashaIncludedGlobal &&` looks redundant.
> It's not effect to the result by omit the double checking.
> 
> btw, I got an error saving file on scripts/download_charaimages_from_wiki.py
> the split separator is "=" ?

Yeah it is redundant it is a leftover from how I thought it up.
1st case is no akasha weapon included
2nd case can't happen, local can only be true if global is true.
3nd case is a akasha weapon is already included but not this akasha weapon so you can't grant its efects to the character
4th case is a akasha weapon is already included and it is this one so you are allowed to grant it.

Old image download scripts are gone, checked your new one, added a progress bar and updated the readme :D

scripts/download_image.py
```diff
@@ -79,14 +90,20 @@ def main(target='arm', site='wiki', save_dir=None, dry_run=False):
...
+        for line in f:
+            num_lines += 1
+
     with open(filename, encoding="utf-8") as stream:
-        for url in _readlines(stream):
-            save_file = os.path.join(save_dir, url.split(separator)[-1])
+        for line, url in enumerate(_readlines(stream)):
+            file = url.split(separator)[-1]
+            progress(line, num_lines, status=os.path.abspath(os.path.join(save_dir, file)))
+            save_file = os.path.join(save_dir, file)
```
kei-gbf: enumerate() return the first number zero, is it ok for line count?
If not then enumerate(..., start=1) option for the initial number.
num_lines = -1 can be zero start.

"file" variable name
Python 2.x had "file" type object which open() returns.
Python editor/IDE may show "file" as keyword
like this in github code highlight shown. I suggest change the variable name
"name" or "save_name" ... whatever (note: "filename" is already used outer scope)

kei-gbf:
@Groovman 

thanks for the explaining. and nice progress bar works :+1:
fancy output is looks good for user. however
I have another use-case which want to suppress that output.

in linux work on a shell/terminal
there is "script" shell command to save all output into a file.
it's used for an error reporting, working log etc.

In this case, Users are not watching the progress in real-time
they run task and watch the result output later.
so then simple plain text output is better for this case.

I will add options later. Default behavior progress bar is ok as your wrote.
my option is for the different use-case.

kei-gbf:
@Groovman 
I updated the download script https://gist.github.com/kei-gbf/8e878356e3ddee7476d026ec437343eb

a bit edited inside progress function
bar_len and %20s format for terminal width 80 characters.

"status" is too long, and if shorter strings was over written
e.g.
"0000_note.png\r"
"0001.png\r"

will show "0001.pnge.png". format padding number can fill this by spaces.

Groovman:
So with Shiva and Kumbhira support skills is basically everything in what I still had to do in the past that I skipped.

Grauph-Delta is a byproduct of implementing chainforce from windhose sword.

kagekiyo7:
@Groovman
propose "Lyria 25% ougi cap up"
限定の奥義ダメージ(25%)(シエテ) → 奥義ダメージ上限25%UP(ルリア) 

propose "New Characters Uncaps for Patch 2019-02-27"

    "normalOntyou": {
        "M": [17.0, 17.0, 17.0, 17.0, 17.0, 17.0, 17.0, 17.0, 17.0, 17.0, 17.0, 17.0, 17.0, 17.0, 17.0]

https://xn--bck3aza1a2if6kra4ee0hf.gamewith.jp/article/show/143931

Groovman:
> @Groovman
> Thank you for the update.
> 
> propose "Lyria 25% ougi cap up"
> name 限定の奥義ダメージ(25%)(シエテ) → 奥義ダメージ上限25%UP(ルリア/ロボミ)
> is its correct ougi limit 2,150,000÷1,685,000(substance Ougi limit) = 27.596439169%UP ?
> https://docs.google.com/spreadsheets/d/1kea2IL6wLNbw4RNUcrrxMTpoIdlXU13pYOzBXjgoBbs/edit#gid=199555968
> 
> propose "New Characters Uncaps for Patch 2019-02-27"
> 
> ```
> "normalOntyou": {
>     "M": [17.0, 17.0, 17.0, 17.0, 17.0, 17.0, 17.0, 17.0, 17.0, 17.0, 17.0, 17.0, 17.0, 17.0, 17.0]
> ```
> https://xn--bck3aza1a2if6kra4ee0hf.gamewith.jp/article/show/143931

I can add Robomi. The ougi cap up for characters as it is currently is just a bandage. It is just to add it into the equation instead of leaving it out, it would be better to add fixed values into the character templates and use them instead of using crude scale numbers now to model it. I just did it to Eternals because it was quick and easy fix and basically Siete 5* support abvility gave a template.

normalOntyou is the debuff resistant that is unknow, I just left the bahamut debuff res numbers in till confirmed. The 17.0 are hp value which are the default normal hp medium up numbers. Gamewith, eng wiki and so on sometimes have different values +-1 depending on where the info came from. I mostly trust the eng wiki and look at patterns in the numbers.

kagekiyo7:
Thank you for update.

Chrysaor
DA=37.0% TA=3.0%
If it be Companion Weapon, rise DA20% and Attack Up(Normal)20%.
I think Chrysaor should divide into two.
クリュサオル(CW) and クリュサオル
http://hiroxgbf.blog.fc2.com/blog-entry-45.html
http://gbf-verification.blog.jp/archives/17104389.html

Groovman:
> Thank you for the update.
> 
> Chrysaor
> DA=37.0% TA=3.0%
> If it be Companion Weapon, rise DA20% and Attack Up(Normal)20%.
> I think Chrysaor should divide into two.
> クリュサオル(CW) and クリュサオル
> http://hiroxgbf.blog.fc2.com/blog-entry-45.html
> http://gbf-verification.blog.jp/archives/17104389.html

Done :)

kei-gbf:
Here is an idea for 3rd support ability implementation. (code clean-up topic)
Motocal uses Babel, we can use ES6 Generator (function keyword with asterisk) and
for-of loop.

```JSX
function* eachSupport(chara) {
    for (let key of ["support", "support2", "support3"]) {
        if (typeof chara[key] === 'undefined') {
            continue; // Data maybe broken.
        }

        if (chara[key] == 'none') {
            continue; // Safe for skip
        }

        if (typeof supportAbilities[chara[key]] === 'undefined') {
            console.error("undefined support ability ID:", chara[key]);
            continue;
        }

        yield supportAbilities[chara[key]];
    }
}
```

and how to use the generator:

- recalcCharaHaisui 
  - `for (let support of eachSupport(chara[ch])) {`  * snip
- treatSupportAbility
  - `for (let support of eachSupport(totals[key])) {` * snip


The undefined guard will fix #138 issue.
That report said Screen white out on load an old data.
Because of support ability key was changed then data compatibility broken.
This change can protect that white out situation, and report error instead.

----
Babel will translate this code to vanilla JavaScript, no worry about browser support the syntax.
I worry is
- [ ] IDE/Editor support the ES6 or JSX syntax?

Groovman:
> Here is an idea for 3rd support ability implementation. (code clean-up topic)
> Motocal uses Babel, we can use ES6 Generator (function keyword with asterisk) and
> for-of loop.
> 
> ```js-jsx
> function* eachSupport(chara) {
>     for (let key of ["support", "support2", "support3"]) {
>         if (typeof chara[key] === 'undefined') {
>             continue; // Data maybe broken.
>         }
> 
>         if (chara[key] == 'none') {
>             continue; // Safe for skip
>         }
> 
>         if (typeof supportAbilities[chara[key]] === 'undefined') {
>             console.error("unknown support ability ID:", chara[key]);
>             continue;
>         }
> 
>         yield supportAbilities[chara[key]];
>     }
> }
> ```
> 
> Generator usage:
> 
> * in recalcCharaHaisui function
>   
>   * `for (let support of eachSupport(chara[ch])) {`  * snip
> * in treatSupportAbility function
>   
>   * `for (let support of eachSupport(totals[key])) {` * snip
> 
> The undefined guard will fix #138 issue.
> That report said Screen white out on load an old data.
> Because of support ability key was changed then data compatibility broken.
> This change can protect that white out situation, and report error instead.
> 
> Babel will translate this code to vanilla JavaScript, no worry about browser support the syntax.
> I worry is
> 
> * [ ]  IDE/Editor support the ES6 or JSX syntax?

I recommend Webstorm, it is not free, but you can get the Jetbrains toolbox and install the Early Access versions. I use it at work for multiple angular projects and for updating motocal. It is the best web IDE I know and support most of the newest web stuff.

kei-gbf:
@Groovman 
Thank you, I worried here was if this require you to change IDE/Editor settings.
modern IDE is ok, they supports ES6 well.

This is not so high priority topics. let me just explain

Pros:
  - code extendability, if characters got "support4" in future, just add the key to the array of "for-of"
    like:  for (let key of ["support", "support2", "support3", "support4"]) ...
  - checking undefined field, allows removing unused support3 field.
    the most of characters has support3: none, this field can be optional.
    - reduce charaData.json file size. (although not much transfer size if using http compression)
    - #138 is another issue, it's for "unknown" type support ability key.
      the field is exists, but the key is not found in supportAbilities assoc array.
      - in current code, `if (support.type == "none") continue;` but `support` was undefined
Cons:
  - An old JavaScript syntax highlight may not support ES6 "generator", "yield", "for-let-of" ...
  - mixing for-in and for-of style bring confuse
    - no need to re-write all for-in, that has many risks.

Groovman:
Opus weapons with filler values and merged some branches.

kagekiyo7:
@Groovman 
Thanks a lot!

競合回避の為に元カレ本家ではなく、Groovmanさんの方にPRを送信したので代わりにこちらで変更点を記述します。
・連撃のスキル効果量修正
http://hiroxgbf.blog.fc2.com/blog-entry-7.html
・クリティカルのスキル効果量修正
https://twitter.com/i/moments/1040200666260459520
・保存タブのスキル性能、参考文献更新
・軍神のDA部分を新たに設定
https://twitter.com/KST33166653/status/1042911871039299584
・オメガ武器の種類、ガフスキーに説明を追加 (選択式スキルの場合書いてあったほうが便利かなと思ったので)
・効果ターンがあるスキルに(永続)書きの追加
・誤字の修正

Groovman:
I have all your githubs in my webstorm, I can just fetch and run merge on your branches, just work mhow you want I can adapt.

kagekiyo7:
I don't know about Webstorm. If I created new branche, how do I inform you of it?


A problem is encountered in this pull request now… Load Browser Data List, then occur whiteout.

Groovman:
I always check the network graph from time to time: https://github.com/MotocalDevelopers/motocal/network

Chain Force works for me, you need to set chain number under player to something greater than 1, because with just MC in chain you get no chain burst.

kagekiyo7:
Okay. I don't need to send you pull request, do I?

Oh! I forgot it. Sorry, I confirmed that Chain Force is working.

I read your "Opus Weapons".
not 振り子キー but ペンデュラム.
and, you add effect of skill(HP(L)) it later?

Groovman:
Kamui aka Primancy has the same HP mod as ATK mod so you have to reuse magna or normal for hp mod. No normalHP or mangaHP L needed.

kei-gbf:
@Groovman 

thank you a lot update!

I made PR with [WIP] tag is Work In Progress
I had some experimental changes in that commits.
I will remove the [WIP] tag when the status is ok to be merged.
(I worry conflict happen in merge, because I may use git rebase for cleanup commit logs.)

kagekiyo7:
@Groovman
I got it.

----
The addition of 3rd skill slot now causes a problem that past party data can not be read.
Am I not recreating party data?

Groovman:
@kagekiyo7

I look briefly at the save a load functions and the problem is mainly with how the save and load functions work. 

Disclaimer massive nerding on how saving and loading data works :)
Saving serializes the current state of the app into a Json, loading deserializes the saved json and replaces the current state. When I add fields to weapons, or kei-gbf adds a new field for defense debuffs, how the state looks changes and when a state before those changes gets loaded the the newly added values are missing aka are undefined. That makes the app go dead, you could add a bunch of checks for undefined but that would just cause extra work. A proper solution is to write a function that resets the state basically an initializon function for the app does the trick can be used for both loading the app for the first time and resetting the app on load. Then you need to iterate over the saved data and assign values to the reset state that are in the saved state and in the app state and do the same for character objects and weapon objects by creating a new object and filling it and then adding it to the app state. So basically the load function need to be made a little bit more intelligent than just saving and overwriting the current state.

Tl;dr: when loading reset the current state of the app and load the data into the clean state by iterating over the saved data instead of replacing the current state with the saved data.

Hope it is understandable what I mean. Writing on my phone on the go so can’t proofread much :)

kagekiyo7:
I got it.
I hope that someone else kindly write function that replace the saved data.

----
Report:Opus Weapons

Slv16~20 of Primacy and Stamina and Enmity of normal and magna is not working.

Primancy effect amout size: 18.4, 18.8, 19.2, 19.6, 20.0 (Slv16,17,18,19,20)
http://gbf-verification.blog.jp/archives/17666527.html

Enmity amout size =（2*x^2-5*x+3）* 基礎倍率(Skill value) 
（ x(0.0≦x≦1.0) =  Current HP / Max HP  ）
https://twitter.com/Hecate_mk2/status/981779724862029824

Skill value=
https://gbf.wiki/Weapon_Skills#ATK_Up_as_HP_lowers
(Gauph Key of Zeal = 10%)

kagekiyo7:
Report:  Bug of "HPDOWN15%" of support ability

・Applied to the first party but not to the second and later.
image

・The effect disappears when you add a weapon.
image

gin0606:
@Groovman 
Can we merge this pull-request?
if we can merge this PR, plz resolve conflict :)

kagekiyo7:
thank you for the update!

----
Report: Character and Weapons for Patch 2019-04-15
not 通常ナ本質(中) but 通常本質(中).
Two or more same element "tenshiShukufukuIII" effect have no effect. (in the game)

----
Slv20 of Sante(M) of Opus Weapons is probably 6.0%.
https://twitter.com/konakatatika/status/1121097405858893824
This is verification of Sante(normal, Slv20) of Agni x Shiva party.
(11.6+12.5+12.1+16.2)/4 = 13.1%
6.0%x 2.2(Agni aura)=13.2% UP
and, also this is a nice round number.

Groovman:
I am working on the last patch and fixing the stuff kagekiyo7 found. Doing a commit and then doing a merge and then merge all of it.

PR #177, opened 2019-05-06
Partial Load feature

kei-gbf:
This patch add dropdown button for partial loading

partial-load

This change makes users manages their save data as Preset.
e.g. Load earth-katana party + Load earth-magna2 weapons.

Save data will become more re-usable.

----

## TODO

- [x] translate button labels
- [ ] CSS (English locale break button group layout)
- [ ] need review about `simulator`

----

__This is an experimental feature, yet__

The partial loading may break user's save data,
If user had clicked the load button by mistake, but there is no Revert.

Please save a backup, then try at your own risk.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -145,9 +145,12 @@
 
     .dataControl .systemButtonGroup {
         width: 100%;
+        text-align: center;
     }
     .dataControl .systemButton {
+        /*
         width: calc(100%/3);
+        */
     }
 
     .tweet .tweetButton {
--- a/src/content.js
+++ b/src/content.js
@@ -669,10 +669,14 @@ var Sys = CreateClass({
                     <Button bsStyle="primary" className="systemButton"
                             onClick={this.onSubmitLoad}>{intl.translate("ブラウザデータ読込", locale)}</Button>
                     <DropdownButton title="&nbsp;" bsStyle="primary" className="systemButton" id="partial-load-menu">
-                        <MenuItem onClick={this.onSubmitPartialLoad.bind(this, 'profile')}>Load Djeeta</MenuItem>
-                        <MenuItem onClick={this.onSubmitPartialLoad.bind(this, 'summon')}>Load Summon</MenuItem>
-                        <MenuItem onClick={this.onSubmitPartialLoad.bind(this, 'chara')}>Load Charactor</MenuItem>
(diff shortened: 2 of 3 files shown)


kei-gbf:
screenshot

In progress, how to remove the space.

kei-gbf:
I will withdraw this PR. The concept was duplicated with #201 and it's better approach.
After I implemented this feature, I noticed several issues.

1. Server may store an uncompleted data,
  if users made the partial data then save to server.
  I assumed users save it to browser. at least everyone have to make the initial data.
 #201 approach, server admin can provide the initial data and users can share use.
 This approach can't provide the "share"

2. maintenance cost
no need to provide same concept by two ways. when #201 was implemented, this will be an unnecessary feature.  

I was going to tell this in #201, add UI code in current code base increase the legacy code.
Motocal use react bootstrap old version, when we need update the version, the cost is increased.
in #201 approach, I assume external program/UI can provide the preset.
e.g 3rd party tool, they can start new repo with new libraries.


3. "simulator" section in the state (save data)
I had not use damage simulator,
I am not sure if this patch had side effect to the simulator feature.
If this PR will be merged, need much tests about simulator.

There are no E2E tests now, test manually is hard to do.

----
I will still keep this PR until #201 close. (for discussion if someone had ideas)
There is also an advantage, users can make custom initial data.
instead of server provide preset.

I will show #201 contains the most of this PR aimed then this PR will be unnecessary.


このPRは撤回予定です。理由は主に３つ、実装してみてわかった事。（要約）
1. サーバに不完全なデータが保存される可能性ある。2. 保守性の問題。
3. ダメージシミュレータ関連に影響がありそうだけどテストが困難。

議論の為に #201 が閉じるまで残しますが、
このPRで提供するはずだった機能は、#201 で(別の形で)実装する予定です。

kei-gbf:
Other comparison #177 and #201
Think how many steps users takes

- #171 case User select
  1. Djeeta profile/load profile
  2. Magna summons/load summons
  3. Magna2 weapons/load weapons
  4. Katana chara/load characters
- #201 case (future scenario)
  Although #201 does not provide the UI directly, but make it possible 3rd party tool can provide like this:
  Users can click/select  Element/Magna summons/Magna2 weapons/Katana chara then click Load

kei-gbf:
I will restart this topic in #201 soon, close here.

PR #185, opened 2019-05-08
Download Image Module Optimizations

OrkunKocyigit:
Current download_image module does not obey PEP8 and PythonZen recommendations and handles every action in the main thread. This pull requests aims to:

- Migrate parameter parsing to optparse library which handles all parameter ordering, help text and validations automatically. With this update user no longer has to parameter order to use script.
- Externalize some of the internal parameters for public use.
- Move url connections to background threads and implementation of ThreadExecutorPool which allows multithreading for better performance.


kei-gbf:
multithread support is looks nice!
`optparse` was in my to-do (never-do) list :)

I'm not running the script yet. this is just 1st impression review.
not running pep8 check right now though, I know Zen well Python is my main language.

- [ ] no need to expand all to module's globals, I suggest use module as namespace.
  - [ ] Problem: `import download_image; dir(download_image) will show those names all.
  - [ ] `info`, 'error' those short names, we may use in local scope.
  - [ ] `logging.` module prefix make it easy to search logging line in source by grep /logging/
  - [ ] import a lots from `os.stat` not do in globals, but you can do inside function scope.
    - [ ] `join' is not sure `os.path.join` or `str.join`.
      can name alias `from os.path import join as path_join` but os.path.join is good enough.
- [ ] no need to be so strict about _url, _path
  there are not exposed to outside. no conflict in the scope there are used.
- [ ] def download_image(durl, dpath): can be `url`, `path` ?
- [ ] def parse_file(stream, _readlines=partial(map, str.rstrip)):
  - `_stream` was not internal use. name can be `stream`
  - `_separator=separator` it's safe to remove. separator is a string and
     python has closure (you can use the outer scope object, they are stored into function object runtime)
  - `_readlines` is only used once in that function, she can move there.
    default variable are run only once at compile time
- [ ] `argvs` ? no need to think naming conflict if use sys.argv and def main(argv):

kei-gbf:
here is reported by `pep8` check script.

```
download_image.py:51:80: E501 line too long (87 > 79 characters)
download_image.py:134:80: E501 line too long (117 > 79 characters)
download_image.py:139:80: E501 line too long (116 > 79 characters)
download_image.py:140:80: E501 line too long (85 > 79 characters)
download_image.py:142:80: E501 line too long (87 > 79 characters)
download_image.py:147:80: E501 line too long (116 > 79 characters)
download_image.py:148:80: E501 line too long (102 > 79 characters)
download_image.py:149:80: E501 line too long (82 > 79 characters)
download_image.py:151:80: E501 line too long (95 > 79 characters)
download_image.py:152:80: E501 line too long (94 > 79 characters)
download_image.py:154:80: E501 line too long (81 > 79 characters)
```

there are options:

- set max-line-length longer
- or ignore the error code if you do not need.
  - in a comment or config file for pep8 checker.

kei-gbf:
@OrkunKocyigit 

did you run something auto formatting tool?

```python
choices=["wiki", "game"])
```

choices can be single line, they are short enough.

```python
"--target", "-t"
```

is acceptable. it's same semantic option.
no need to separate so often if it was short enough.

----
```python
future_to_image = {
                    executor.submit(
                        download_image,
                        url,
                        path): (
                        url,
                        path) for (
                        url,
                        path) in items}
```

is hard to read here. length too long is for width
height also an important point. display 1 function without scroll.
(expect long main function usually)

they can be two lines. I have an idea using functools.partial here.

```python
submit = functools.partial(executor.submit, download_image)
future_to_image = {submit(url, path): (url, path) for (url, path) in items}
```

I am not testing this code, sorry if this did not work with multithread.

----
string format too. no need separate all parameters.
% operator  may be now suggested string format method.
(I am not sure this, not pep8 or style guide)

----
- [ ] logging as log is .... there is `log` function in logging module, the name became log.log 
      or readers may think it's logging.log

OrkunKocyigit:
@kei-gbf 
I use autopep8 in order to conform PEP8 standarts and check it with pycodestyle to verify it. My configuration was set max line size to 120 so i fixed it by reducing it to 79.

I fixed most issues on your first comment, I'll check your current one right now.

Cheers

OrkunKocyigit:
I check it up on notepad and yeah it is pretty messy :(. I'll fix formatting by hand

kei-gbf:
```python
import os.path as op
```

no, please do not do this. reader need to search what is the `op`

```diff
- _readlines = partial(map, str.rstrip)
+ read_lines = functools.partial(map, str.rstrip)
```

the `_` prefix was for it's module private.
the prefix has a meaning in Python.`from download_image import *` will not import _ prefixed names.

kei-gbf:
Ahh, `notepad` it's hard to edit Python code
take care about indent, no mix TAB character, ident must be soft spaces

> I fixed most issues on your first comment, I'll check your current one right now.

I did not update after the first commit. it was for basic idea.

OrkunKocyigit:
Latest revision should be fine I think, I'll try to run some benchmarks with timeit.

OrkunKocyigit:
New script - All weapon pictures with 10 workers: 63.1349308 (All sha1 matches with old script)
New script - All weapon pictures with 1 worker: 528.408152 (All sha1 matches with old script)
Old script - All weapon pictures: 182.9112399
New script - All character pictures with 20 workers: 100.08886689999997 (All sha1 matches with old script)
New script - All character pictures with 1 worker: 589.5705985999999 (All sha1 matches with old script)
Old script - All character pictures: 192.7906309

kei-gbf:
Optional long name was a good practice, even no one use.
optparse generate usage document from options.
long name can describe what option.

```
> python3 download_image.py -h
Usage: 
    Download image file in URL list.

    Options: [--target arm] [--site wiki] [--save_dir path] [-d] [--workers=1]
      * -t: (arm|chara)
      * -s: (wiki|game)
      * --save_dir: (./imgs|./charaimgs)
      * -d: Prints url list without download.
      * -w: Number of threads to download images.
      * -r: (progress|plain)
      * -o: Redownloads all images even if it exists
```

I've noticed need to update main doc.

`    Options: [--target arm] [--site wiki] [--save_dir path] [-d] [--workers=1]`

list of common option names used in unix commands:

- -d for directory (--save-dir)
- -d dry run  can be  -q --quiet
- -o overwrite can be -f --force
  `o` is often used for output option.

kei-gbf:
@OrkunKocyigit 

which environment did you run the code?
I think there is an issue main thread is gone before all tasks completed.

- executor.shutdown() to wait tasks completed in main theread
- no need to call future.result() ?
- I got 503 error service temporary unavailable ... too much access?
  I worry now site may block the access from download script.
- too much worker but http server may limit number of connections

OrkunKocyigit:
I use python 3.7.2 with venv on game server.
image

kei-gbf:
I misunderstood about optparse option.
current code was add_help_option=False and not given sys.argv[0]

Can you try
main(sys.argv) and add_help_option=True then python3 image_download.py -h
optparse generate those options document, no need to add in main.__doc__

OrkunKocyigit:
Here is the results for that:
Usage: download_image.py [options]

Options:
  -h, --help            show this help message and exit
  -t TARGET, --target=TARGET
  -s SITE, --site=SITE  
  -d SAVE_DIR, --directory=SAVE_DIR
  -q, --quiet           
  -w WORKERS, --workers=WORKERS
  -r REPORTER, --reporter=REPORTER
  -f, --force           

It is possible to add messages to options but gotta add them one by one with help="Message" argument to each option like this:
parser.add_option('--force', '-f', action='store_true', dest="overwrite", help="Redownloads all images even if it exists")

kei-gbf:
my environment win10/py3.7 on bash terminal (git for windows)
the new script finished without error. but did not download all files.
run the script second time, continue to download next.

> executor.shutdown() to wait tasks completed in main theread

this is self reply. I noticed executor is used in 'with' statement.
so shutdown is called.

OrkunKocyigit:
image
image

kei-gbf:
```
-t TARGET, --target=TARGET
-s SITE, --site=SITE
```
Users may want to see help there.

help="Data to be downloaded (arm|chara)",
help="Source to download from (wiki|game)",

hmm ... long name is now a bit redundant?
because dest naming also describe the same twice.
but it's ok acceptable if not too long.

----

Yes, same console. my prompt also shows MINGW64
and my result was 29 files.

dir property

OrkunKocyigit:
I added try-except block to download image, can you try again and see if you get any errors printed on console?

kei-gbf:
you can try open this PR in gitpod sandbox. same happen
https://gitpod.io/#https://github.com/MotocalDevelopers/motocal/pull/185

NOTE: I needed to edit "force" -> "--force"

```
> cd scripts
> python3 download_image.py arm
```

キャプチャ

report shown 917 lines but files are less.

> I added try-except block to download image, can you try again and see if you get any errors printed on console?

There was no errors. the 503 error was when I try run the script many times.

kei-gbf:
```python
def main(argv):
    """
    Download image file in URL list.

    Options: [-t arm] [-s wiki] [-d path] [-q] [-w=10] [-r=progress] [-f]
      * --target -t: Data to be downloaded (arm|chara)
      * --site -s: Source to download from (wiki|game)
      * --directory -d: Save directory (./imgs|./charaimgs)
      * --quiet -q: Prints url list without download.
      * --workers -w: Number of threads to download images.
      * --reporter -r: Type of reporter (progress|plain)
      * --force -f: Redownloads all images even if it exists
    """
```

I like this simple and good enough.
but an issue is when change the parameter
maintainer must update the document together.

```python

def main():
    """
    Download image file in URL list.
    """

parser = OptionParser(usage=main.__doc__)

main(sys.argv)
```

then optparse can generate usage like unix command format.

```
Usage: download_image.py [options]

Options:
  -h, --help            show this help message and exit
  -t TARGET, --target=TARGET
                        Data to be downloaded (arm|chara)
  -s SITE, --site=SITE  Source to download from (wiki|game)
  -d SAVE_DIR, --directory=SAVE_DIR
                        Save directory (./imgs|./charaimgs)
  -q, --quiet           Prints url list without download.
  -w WORKERS, --workers=WORKERS
                        Number of threads to download images.
  -r REPORTER, --reporter=REPORTER
                        Type of reporter (progress|plain)
  -f, --force           Redownloads all images even if it exists
```

kei-gbf:
now I got 503 error,
too many connections. access from the script is like an attack for the service.
worker should be few number?

OrkunKocyigit:
I added a new way to save web results into a file. From what I understand you can download the files but thread is killed before it can save it. This new way in theory should eliminate that.
I also ported OptionParser to ArgumentParser as that library seems to provide linux style usage documentation.

kei-gbf:
> From what I understand you can download the files but thread is killed before it can save it. 

not exactly.  I can download but seem the future task are not even launched,
because main thread are killed before... yes

OrkunKocyigit:
ArgParser documentation example:
(motocal) C:\Users\Jongaros\WebstormProjects\motocal\scripts>python download_image.py -h
usage: download_image.py [-h] [-t {arm,chara}] [-s {wiki,game}] [-d DIRECTORY]
                         [-q] [-w WORKERS] [-r {progress,plain}] [-f]

optional arguments:
  -h, --help            show this help message and exit
  -t {arm,chara}, --target {arm,chara}
  -s {wiki,game}, --site {wiki,game}
  -d DIRECTORY, --directory DIRECTORY
  -q, --quiet
  -w WORKERS, --workers WORKERS
  -r {progress,plain}, --reporter {progress,plain}
  -f, --force

kei-gbf:
usage looks nice. :+1:
coding style review is almost done.

next topics are:

1. Thread issue, main thread is gone before task finish.
2. Server 503 error because too many accesses.
  we can do nothings for this, just take an interval.
  it's not so good to test with real server many times.  

give me a time, I can join the debug.

OrkunKocyigit:
Wiki image fetching broken due to links in the file was using AngularJS to embed specific div objects. I fixed it by writing converter to that actually creates direct links to wiki resources.

    def transform_wiki_url(file_name):
        url = r'http://gbf-wiki.com/attach2/%s_%s.png'
        return url % ('img'.encode('utf-8').hex().upper(),
                      file_name.encode('utf-8').hex().upper())

It would be faster to remove this adapter and convert links in the files in this format manually but this would do for current situation.

kei-gbf:
we need update README ? (about the script usage)

Can you add thise commands into package.json scripts section
```json
"download:arm": "python3 ./scripts/download_image.py arm",
"download:chara": "python3 ./scripts/download_image.py chara",
```

NOTE: `python3` is ok here, it assumes command exists in PATH.

I added Edit CodeSandbox button on #179 
This make it possible run the script on CodeSandbox
(CodeSandbox does not have terminal)

kei-gbf:
Yeh! got 917 files now

download result

OrkunKocyigit:
Only potential solution I can think of for server errors is, if you receive any errors, wait 500ms before trying again and if you fail 3 times in a row, give up.

kei-gbf:
Thank you! download:arm worked on CodeSandbox too.

```python
#!/usr/bin/env python3
"""
Motocal Image download script.
Usage: python download_image.py arm
       python download_image.py chara
Debug e.g: python download_image.py arm wiki tmp -d plain
```

- -t is required ?
- README also has same usage.
  - drop this info? or README can tell `npm run download:arm`
- s/python/python3/
- Debug line is now obsolated. (that was for dry run, plain reporter)

OrkunKocyigit:
> 
> 
> Thank you! download:arm worked on CodeSandbox too.
> 
> ```python
> #!/usr/bin/env python3
> """
> Motocal Image download script.
> Usage: python download_image.py arm
>        python download_image.py chara
> Debug e.g: python download_image.py arm wiki tmp -d plain
> ```
> 
>     * -t is required ?
> 
>     * README also has same usage.
>       
>       * drop this info? or README can tell `npm run download:arm`
> 
>     * s/python/python3/
> 
>     * Debug line is now obsolated. (that was for dry run, plain reporter)

- Yep, argparser library depends on argument names.
- I only added user would require Python 3.7+ to run this script. I can add that but some users might want to run scripts manually as well.

I also push retry update in case of server throws an error other than 404 for obvious reasons.
    def download_image(url, path):
        success = False
        error_count = 0
        while not success:
            if not options.quiet:
                try:
                    with urlopen(url) as response, open(path, mode="wb") as image_file:
                        copyfileobj(response, image_file)
                        response.close()
                        success = True
                        break
                except HTTPError as error:
                    if error.code == 404:
                        print("Bad Url %s at path %s" % (url, path))
                        error_count = 3
                    else:
                        error_count += 1
                    if error_count >= 3:
                        break
                    time.sleep(0.5)
        return success

Following url&files are probably invalid. I don't know who maintains them.
Bad Url http://gbf-wiki.com/attach2/696D67_333034303035373030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040057000_03.png
[==-------------------------------------------] 3.7% ...   3040057000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333034303034353030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040045000_03.png
[===------------------------------------------] 7.4% ...   3040045000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333034303135333030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040153000_03.png
[=====----------------------------------------] 11.1% ...   3040153000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333034303036333030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040063000_03.png
[=======--------------------------------------] 14.8% ...   3040063000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333034303034393030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040049000_03.png
[========-------------------------------------] 18.5% ...   3040049000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333034303132323030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040122000_03.png
[==========-----------------------------------] 22.2% ...   3040122000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333034303035323030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040052000_03.png
[============---------------------------------] 25.9% ...   3040052000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333034303032393030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040029000_03.png
[=============--------------------------------] 29.6% ...   3040029000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333034303037313030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040071000_03.png
[===============------------------------------] 33.3% ...   3040071000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333034303138333030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040183000_03.png
[=================----------------------------] 37.0% ...   3040183000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333034303031323030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040012000_03.png
[==================---------------------------] 40.7% ...   3040012000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333034303031393030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040019000_03.png
[====================-------------------------] 44.4% ...   3040019000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333034303030383030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040008000_03.png
[======================-----------------------] 48.1% ...   3040008000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333034303030373030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040007000_03.png
[=======================----------------------] 51.9% ...   3040007000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333034303032313030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040021000_03.png
[=========================--------------------] 55.6% ...   3040021000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333034303030313030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040001000_03.png
[===========================------------------] 59.3% ...   3040001000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333033303030383030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3030008000_03.png
[============================-----------------] 63.0% ...   3030008000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333033303030373030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3030007000_03.png
[==============================---------------] 66.7% ...   3030007000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333033303035313030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3030051000_03.png
[================================-------------] 70.4% ...   3030051000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333033303035343030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3030054000_03.png
[=================================------------] 74.1% ...   3030054000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333033303036313030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3030061000_03.png
[===================================----------] 77.8% ...   3030061000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333033303036353030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3030065000_03.png
[=====================================--------] 81.5% ...   3030065000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333033303037313030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3030071000_03.png
[======================================-------] 85.2% ...   3030071000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333033303131313030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3030111000_03.png
[========================================-----] 88.9% ...   3030111000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333033303038383030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3030088000_03.png
[==========================================---] 92.6% ...   3030088000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333033303137323030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3030172000_03.png
[===========================================--] 96.3% ...   3030172000_03.png
Bad Url http://gbf-wiki.com/attach2/696D67_333033303039323030305F30332E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3030092000_03.png
[=============================================] 100.0% ...   3030092000_03.png

kei-gbf:
Oh, error handling. personally I do not feel much work there.
It's hard to test with real server and
It only once in setup time to download such many files.

keep download script simple as possible.
users can run the script again if it was failed. (if it was not automated deploy)

Is it ok to use time.sleep in multithread code?

kei-gbf:
missing image link or url changed ? I may have the patch, I made before ...

kei-gbf:
I have a patch about these 26 chara
I forgot this I used in local but did not reported.
(because no need to download all again, only new files)

 - [ ] ガンダゴウザ
 - [ ] アルベール
 - [ ] バザラガ(闇)
 - [ ] レナ
 - [ ] アルルメイヤ (土)
 - [ ] メーテラ (風)
 - [ ] サルナーン (光)
 - [ ] アルタイル
 - [ ] オイゲン
 - [ ] ラカム
 - [ ] 城ヶ崎美嘉
 - [ ] リタ
 - [ ] エジェリー
 - [ ] アーミラ(SR)
 - [ ] 双葉杏
 - [ ] 城ヶ崎莉嘉
 - [ ] ウィル
 - [ ] ヤイア
 - [ ] ミュオン
 - [ ] 桜内梨子＆高海千歌＆渡辺 曜
 - [ ] サンダルフォン
 - [ ] アンチラ
 - [ ] スカーサハ
 - [ ] ナルメア (闇)
 - [ ] ヴァンピィ
 - [ ] シルヴァ (水)

kei-gbf:
27 files ? my list had only 26. coz it was an old work.

kei-gbf:
@OrkunKocyigit 

I was going to send PR to your branch update README-ja
but github did not ask me open PR
seem I had edited README-ja.md directly on github site... no permission!?
sorry if it break the sync of your workspace to remote.

kei-gbf:
> ArgParser

the `target` can be positional argument? then can remove the -t

kei-gbf:
```python
def download_image(url, path, _retry_count=3):
    if options.quiet:
        return

    for _ in range(_retry_count):
        try:
            with urlopen(url) as response, open(path, mode="wb") as image_file:
                copyfileobj(response, image_file)
                response.close()
                return
        except HTTPError as error:
            if error.code == 404:
                print("Bad Url %s at path %s" % (url, path))
                break
            time.sleep(0.5)
```
I did not test this code. just showing a refactoring approach.
"success" and "error_count" flag can be removed. decrease the nested block

> Flat is better than nested.

OrkunKocyigit:
> missing image link or url changed ? I may have the patch, I made before ...

I added missing file info to your PR.



> @OrkunKocyigit
> 
> I was going to send PR to your branch update README-ja
> (becoz I want this PR includes that)
> but github did not ask me open PR
> seem I had edited README-ja.md directly on github site... no permission!?
> sorry if it break the sync of your workspace to remote.

Wiki, issues and readme files are open to public so you are welcome to edit as it requires. Thanks for editing Japanese files.

OrkunKocyigit:
@kei-gbf 
I updated package.json scripts and converted target as positional argument.
You can also pass parameters to python script via npm (I was not aware of such use case until googled it)
For example:
npm run download:arm -- -h

Updating documentation to reflect that seem like a better approach. I'll be uploading updated English Readme

OrkunKocyigit:
> ```python
> def download_image(url, path, _retry_count=3):
>     if options.quiet:
>         return
> 
>     for _ in range(_retry_count):
>         try:
>             with urlopen(url) as response, open(path, mode="wb") as image_file:
>                 copyfileobj(response, image_file)
>                 response.close()  # close() maybe called by with statement ?
>                 return True
>         except HTTPError as error:
>             if error.code == 404:
>                 print("Bad Url %s at path %s" % (url, path))
>                 break
>             time.sleep(0.5)
> ```
> 
> I did not test this code. just showing a refactoring approach.
> "success" and "error_count" flag can be removed. decrease the nested block
> 
> > Flat is better than nested.

Refactored download_image:

>     def download_image(url, path, _retry_count=3, _timeout=1000):
>         for _ in range(_retry_count):
>             try:
>                 with urlopen(url, timeout=_timeout) as response, open(path, mode="wb") as image_file:
>                     if not options.quiet:
>                         copyfileobj(response, image_file)
>                     return True
>             except HTTPError as error:
>                 if error.code == 404:
>                     print("Bad Url %s at path %s" % (url, path), file=sys.stderr)
>                     break
>                 time.sleep(0.5)
>         return False

kei-gbf:
if not options.quiet: inside open()
does not make empty files if quiet is True ?

btw, my mistake, dry-run was not equal to quiet,
quiet seem disable reporter option.
dry-run is for test/checking what files to download without download in fact.

kei-gbf:
This is NOT change request, a bit off topic.
(not pep8, but coding style guide may mention this)

```python
for url in _read_lines(file):
```

I found "file" variable name.
here is syntax highlight sample:

```python
def test(path, file, a, b, c):
    for i in range(10):   # reserved word: red
        pass
    path = None   # normal variable name
    list = None   # builtins color: blue
    file = None  # orange? like keyword argument
    print("TEST", file=sys.stderr, end='')
```

It's a historical reason, `file` was built-in type in Python2.x
syntax highlighter still treat 'file' something special. 
we avoid to use file as variable name, because of the color shown in Editors/IDE.
but it's still ok to use, as you can see print() function takes file as parameter name.

----

Instead of rename `file` variable name.
iterator/generator abstraction can solve this situation.

```python
def scan_file_for_download_list(iterable):
    for line in iterable: # just for explain test now.
        yield line

## Test code
# _read_lines can go outside of the function
# this change make it easy to write a test, no need mock input stream. e.g)
items = scan_file_for_download_list(["line1", "line2", "line3"])
assert len(list(items)) == 3

## Actual code (generator can remove the list.append)
items = list(scan_file_for_download_list(_read_lines(stream)))
```

please rename `scan_file_for_download_list` if this line hit length >80 :)

OrkunKocyigit:
I can read dry-run and add quiet as a reporter option if that is okay?

There are two problems with using generators instead of lists that I encountered.
- In order to get total item size generator has to be iterated whole once then reset again which causes overhead.
- Generator has to be surrendered by a locking mechanism or generator access has to be through thread safe FIFO mechanism such as Queue to not skip any elements which adds more overhead.

While those can be implemented I am not confident it will bring enough benefits to justify cost.

kei-gbf:
> I can add dry-run and add quiet as a reporter option if that is okay?

Yes, thx a lot works

> There are two problems with using generators instead of lists that I encountered.
> 
> - In order to get total item size generator has to be iterated whole once then reset again which causes overhead.
> 
> - Generator has to be surrendered by a locking mechanism or generator access has to be through thread safe FIFO mechanism such as Queue to not skip any elements which adds more overhead.
> 
> While those can be implemented I am not confident it will bring enough benefits to justify cost.

The info is true about using generator inside threads,
however this is out of worker thread. 

```python
items = list(scan_file_for_download_list(_read_lines(stream)))  # NOTICE: <- list() here
total = len(items)  # thread still can use items, its a list 
```

no much benefits yes. I agree about that.
This will be worth when write tests, reuse the function but no chances.

OrkunKocyigit:
Added new options.
download_image.py -h
usage: download_image.py [-h] [-s {wiki,game}] [-o OUTPUT] [-d] [-w WORKERS]
                         [-r {progress,plain,quiet}] [-f]
                         {arm,chara}

positional arguments:
  {arm,chara}

optional arguments:
  -h, --help            show this help message and exit
  -s {wiki,game}, --site {wiki,game}
  -o OUTPUT, --output OUTPUT
  -d, --dry-run
  -w WORKERS, --workers WORKERS
  -r {progress,plain,quiet}, --reporter {progress,plain,quiet}
  -f, --force

kei-gbf:
thank you, this can add -q option 

```python
parser.add_argument('-q', '--quiet', action='store_const', dest='reporter', const='quiet')
```

```python
with urlopen(url, timeout=_timeout) as response, \
            open(path, mode="wb") as image_file:
    if not options.dry_run:
        copyfileobj(response, image_file)
    return True
```

is now URLOPEN and OPEN file for saving but not write if dry_run was True.
I expect dry-run is for testing. not open url/file then just watch report log.

I like the quiet reporter approach, dry-run also can do same switch function

```python
def do_nothing(*args, **kw):
    pass

task = download_image if not option.dry_run else do_nothing
# executor.submit(task, url, path)
```

OrkunKocyigit:
Hello,

Wouldn't that cause quiet to conflict with reporter option. For example what is the excepted result if I do download arm -q --reporter plain. Should I do plain reporting or do nothing? If those two will conflict I can group them so parser can throw error if both of them send at the same time.

One of the reasons I put dry_run in save file function so that script might be used in order to verify if links in txt files are valid or not. Skipping executor task will make it lose that function.

kei-gbf:
> Wouldn't that cause quiet to conflict with reporter option. For example what is the excepted result if I do download arm -q --reporter plain. Should I do plain reporting or do nothing? If those two will conflict I can group them so parser can throw error if both of them send at the same time.

it's up to argparse implementation, I think usually last option is overridden.

e.g.: -q -r plain -r progress => progress

> One of the reasons I put dry_run in save file function so that script might be used in order to verify if links in txt files are valid or not. Skipping executor task will make it lose that function.

I see, that's different use-case. http HEAD request can do the validation faster?

kei-gbf:
This review for the last part.

- drop future_to_image dictionary, (url, path) is pass via future.result()
- flatten unnecessary nest.

in Python, flat is better than long deep nested code.
unlike other language, Editor can not support highlight end of block.
if it would be long then it's time to separate to another function.

```python
with open(filename, encoding="utf-8", mode='r') as url_list_file:
    # CPU wise copying to list is cheaper as we need to load all items into
    # memory in order to count them anyways
    items = list(scan_file_for_download_list(_read_lines(url_list_file)))
    total = len(items)

# break/return first, this approach is used in Python often,
# do not let indent level go deep.
if total < 1:  
    return # can write single thread code, here.

with ThreadPoolExecutor(max_workers=options.workers) as executor:
    submit = functools.partial(executor.submit, download_image)
    counter = functools.partial(next, itertools.count(start=1))
    for future in as_completed(itertools.starmap(submit, items)):
        report(counter(), total, future.result())
```

this code require two changes:
- download_image() return a tuple (url, path) then catch values by future.result()
- reporter takes (url, path) arguments.
  - no need url-or-path logic in thread code. (different semantics layer)
    simple wrapper can solve it,

minimum sample, how download_image and
reporter function interface change for this.

```python
def download_image(url, path):
    return url, path # future.result() will return this

def _plain_reporter(count, total, result):
    url, _ = result  # from future.result()
```

----

- [x] pep8, code clean up
- [x] options - ArgumentParser
  - [x] quiet option
  - [ ] dry-run
- [x] MultiThread support.
- [x] update README script usage.
- [x] add to package.json for npm run (and yarn run script for CodeSandbox)

OrkunKocyigit:
@kei-gbf 
I mitigated my code to satisfy your requirements. Can you check it?

kei-gbf:
@OrkunKocyigit 

- -v is usually for version option, but Python -v is also not version. it's ok
- validate and dry run  thx
  - they can move to outside of "with open"
    because of now nest level is deep.
- other "options" in function is outer scope variable, this make the function closure.
  I know I mentioned this, but now let me explain the demerit side.
  e.g.case: download_image function code size it's grown now, let move to outside of main.
  -> options have to be parameter. (it's a scope dependency)

NOTE: closure is still useful in short case, and also take many parameters from outer scope.
it does not mean no use closure. case-by-base or trade-off stuff.

I think grouping "options" handle section is better (if it's possible)

- def main
  - parse options
  - handle options
  - main logic (no refer to options anymore)

kei-gbf:
This view point is reduce module level expose
dir(donwload_image) show many entries.

now we can use:
`
     'quiet': _do_nothing,
`

_read_lines ... is only used once, can move to inner function

no need to hide all, but there was unnecessary case.

kei-gbf:
@OrkunKocyigit 

I sent PR to your branch (for showing the code I explained above)

OrkunKocyigit:
Thanks I'll check and split my main method into smaller chunks as well

kei-gbf:
my fault, I did not run checker.

```diff
- _read_lines=functools.partial(map, str.rstrip)
+ _read_lines = functools.partial(map, str.rstrip)
```

I had changed your validation logic, I note in FIXME comment.

kei-gbf:
I need to explain a bit.
what itertools solves.

```python
# problem: the distance between enumerate and start=1
for num, future in enumerate(as_completed(future_to_image), start=1):
    report(num, total, future.result())
```

```python
# then two lines, still line length long,
# Python does not like the long indent spaces.
for num, future in enumerate(as_completed(future_to_image),
                             start=1):
    report(num, total, future.result())
```
remember Zen:, flat is better than nested. avoid the unnecessary iterator nest
itertools.count(start=1) is an idiom to replace enumerate situation.

```python
counter = itertools.count(start=1)
for future in as_completed(future_to_image)
    report(counter(), total, future.result())
```

however, without itertools.count, it's same.
just need to think name for the temporary variable.

```python
completed_futures = as_completed(future_to_image)
for num, future in enumerate(completed_futures, start=1):
    report(num, total, future.result())
```

the iterator nest can be flat though,
good enough if the length was not too long.

iterator nest make debug harder if it was deep nested.
no problem in this simple code, and not so deep.

kei-gbf:
download_image docstring

```diff
- download image (for worker method)
+ download image (for worker thread)
```

or add more details, it was almost same as main's docstring now.

kei-gbf:
validate=False argument for download_image func.
it's my misunderstood. the parameter is not given, it's always False.

options.validate gives method='HEAD'

kei-gbf:
Reporter in OOP readers may assumes
- Reporter
  - ProgressReporter
  - PlainReporter
  - QuietReporter

it's a bit heavy to implement them all for this single script.

I can show small solution.
in Python, function decorator/closure helps bind such dynamic parameter.

```python
def bind_count_total(total, report):
    count = itertools.count(start=1)
    @functools.wraps(report)
    def _func(*args, **kw):
        return report(next(count), total, *args, **kw)
    return _func
```

```python
# usage:
report = bind_count_total(total, _plain_reporter)

report(future.result())
```

It's a decorator, but "total" is decided run time, can't do this.
```python
@functools.partial(bind_count_total, total) # XXX: don't know total yet
def _new_reporter(count, total, message):
    pass
```

----
NOTE: not tested yet

kei-gbf:
where increment self.count ?

_validate_report_function can be function without state (self. inside method)
_validate_report_function(reporter : collable) make the function easy to test.

Accessing code object bring dependency if the function was wrapped by decorator?
lose flexibility of function parameters interface.
and lose portability may not work another python implementation.
there are only 3 reporter, we can check manually. (or write test code can ensure it)

kei-gbf:
Reporter.report added multiline, it's _progress_reporter specific,
This change make composed situations, like
another reporter take message with '\r' etc, such unused or rare cases.
it increase debug/test costs.

There are no parameter for passing multiline from command line, now.
I propose the drop the multiline flag solution. (snippets)

```python
# REPORT_TYPE
`progress-multiline': functools.partial(_progress_reporter, end='\r'),

def _progress_reporter(count, total, message, end=''):
    : 
    print('[%s] %s%s ...%20s' % (bar, percents, '%', status), flush=True, end=end)

# TODO: update comments, readme etc by change the usage
```

kei-gbf:
I think no need Reporter class,
the class does not do much work as the code size.
passing total, count from main thread is ok.

kei-gbf:
One of issue, we are missing unit tests.
it helps think about how program code should be designed.

no test codes, I did not want to manage two or more files,
just small work at the first time.

The original code was just wget and rename onliner shell script.
and download image is task only once in setup/update time.
dostring is good practice, but another opinion it unnecessary and make size large.
I feel now, no need so much resource to develop this single script, (but thx a lot works)

Implement Motocal javaScript worker may be interesting topic?
(calc background can fix motocal freeze sometime when too many weapons)

- [ ] no Report class
- [ ] no need _validate_report_function 
      single test code can test function parameter, without the complex information.

kei-gbf:
doc-test in doc-string, support write a test in single file.

```python
def _plain_reporter(count: int, total: int, message: str):
    """
    >>> _plain_reporter(1, 2, "OK")
    [    1/    2] Download OK
    """
    print("[{:4}/{:4}] {}".format(count, total, message))
```

kei-gbf:
btw, api doc is nice, I did not mean revert the document. tho.
it's up to cases, how often the module is used
this script is now single standalone, no one use as library. (as I know)
so only mainteners read. it's useful if it was a library.

This is also a point, keep the size small, make it easy to maintenance.
instead of a lot documents.

----

```python
    if not validate_source_files(source_location,
                                 filename) or not _create_save_locations(
        options.output, options.target, script_dir):
        return
```

it's a bit tricky for me. 

```python
    if (not validate_source_files(source_location, filename) or
        not _create_save_locations(options, script_dir)):
        return
```

but, no need to move such one-liner,
it just increase line size for pass parameters/return success,
readers have to see the function for the single line. it's overhead.

scan_file_for_download_list need the optimization.

scripts/download_image.py
```diff
@@ -47,13 +48,25 @@ def __init__(self, total: int, report_function: callable, *args, **kwargs):
                                                  **kwargs)
         self._validate_report_function()
 
-    def report(self, message, multiline=True, *args, **kwargs):
-        if not multiline:
+    def report(self, message: str, next_line: bool = True, *args, **kwargs):
```
kei-gbf: message: str here but type check fail, reporter functions message: FutureResult

scripts/download_image.py
```diff
@@ -47,13 +48,25 @@ def __init__(self, total: int, report_function: callable, *args, **kwargs):
...
+        """
+        Prints requested message to the console.
+        :param next_line: whatever reporting should skip a line or not.
+        :type message: str message to print to the console.
+        :return:
+        """
+        if not next_line:
```
kei-gbf: This is _progress_reporter specific code?

This change increase test costs, another reporter also may take the message with next_line.

scripts/download_image.py
```diff
@@ -63,58 +76,73 @@ def _validate_report_function(self):
                 ' total and message arguments')
 
 
-def _progress_reporter(count, total, message):
-    bar_len = 45
+def _progress_reporter(count: int, total: int, message: FutureResult,
```
kei-gbf: naming of `message`
message += '\r' was look like message is string

"future" is better here,
Readers search the name in source, then they will find future.result() where the value was passed from.

scripts/download_image.py
```diff
@@ -123,87 +151,162 @@ def download_image(url, path, method='GET', validate=False,
...
+        options.output, options.target, script_dir):
+        return
 
     def transform_wiki_url(file_name):
         url = r'http://gbf-wiki.com/attach2/%s_%s.png'
         return url % ('img'.encode('utf-8').hex().upper(),
                       file_name.encode('utf-8').hex().upper())
 
-    def scan_file_for_download_list(url_map):
+    def scan_file_for_download_list(url_map: map, site: str, output: str,
```
kei-gbf: map is builtin function in python, not type.
map type is data type in another language, and it's dictionary for python.
typing.Iterable but list is also ok, since it not checked at run time.

scripts/download_image.py
kei-gbf: namedstuple for annotation type check does not work thought, (do not know field types)
new Python has data type instead of namedtupe. that's work for type check.
personally, I feel do not need update python for so unnecessary new feature.
just note the info.

OrkunKocyigit: Updated message to future_result. Removed Reporter class and all of it is methods.

scripts/download_image.py
OrkunKocyigit: Added some logic to all reporters and removed string appending

scripts/download_image.py
OrkunKocyigit: Removed

scripts/download_image.py
OrkunKocyigit: Changed to list

kei-gbf:
several minor clean up.

_read_lines can be read_lines now, its not module's top-level so not exposed.
_max_workers as well.

```diff
- logging.error("Bad Url %s at path %s" % (url, path))
+ logging.error("Bad Url %s at path %s", url, path)
```
there are some using % operator in logging.error

kei-gbf:
I withdraw this:

```python
# abuse of functools
counter = functools.partial(next, itertools.count(start=1))
reporter(counter(), total, future.result())
```

```
# short and simple
counter = itertools.count(start=1)
reporter(next(counter), total, future.result())
```

A point here was flatten enumerate() iterator by itertools.count(). not functools.

OrkunKocyigit:
> 
> 
> @OrkunKocyigit
> I closed #186 because it was incorrect way.
> can you fix the link URL ? I am not sure this info.
> 
> url.replace("_03.png", "_03_full.png") # when it's request to wiki
> and save as _03.png then no needed to change charaData.json

This causes errors:
```
ERROR:root:Bad Url http://gbf-wiki.com/attach2/696D67_333034303035313030305F30335F66756C6C2E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040051000_03.png with error HTTP Error 404: Not Found
ERROR:root:Bad Url http://gbf-wiki.com/attach2/696D67_333034303034333030305F30335F66756C6C2E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040043000_03.png with error HTTP Error 404: Not Found
ERROR:root:Bad Url http://gbf-wiki.com/attach2/696D67_333034303034303030305F30335F66756C6C2E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040040000_03.png with error HTTP Error 404: Not Found
ERROR:root:Bad Url http://gbf-wiki.com/attach2/696D67_333034303032373030305F30335F66756C6C2E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040027000_03.png with error HTTP Error 404: Not Found
ERROR:root:Bad Url http://gbf-wiki.com/attach2/696D67_333034303032333030305F30335F66756C6C2E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040023000_03.png with error HTTP Error 404: Not Found
ERROR:root:Bad Url http://gbf-wiki.com/attach2/696D67_333034303031333030305F30335F66756C6C2E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040013000_03.png with error HTTP Error 404: Not Found
ERROR:root:Bad Url http://gbf-wiki.com/attach2/696D67_333034303031313030305F30335F66756C6C2E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040011000_03.png with error HTTP Error 404: Not Found
ERROR:root:Bad Url http://gbf-wiki.com/attach2/696D67_333034303031303030305F30335F66756C6C2E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040010000_03.png with error HTTP Error 404: Not Found
ERROR:root:Bad Url http://gbf-wiki.com/attach2/696D67_333034303030393030305F30335F66756C6C2E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040009000_03.png with error HTTP Error 404: Not Found
ERROR:root:Bad Url http://gbf-wiki.com/attach2/696D67_333034303030333030305F30335F66756C6C2E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040003000_03.png with error HTTP Error 404: Not Found
ERROR:root:Bad Url http://gbf-wiki.com/attach2/696D67_333034303030353030305F30335F66756C6C2E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3040005000_03.png with error HTTP Error 404: Not Found
ERROR:root:Bad Url http://gbf-wiki.com/attach2/696D67_333033303030333030305F30335F66756C6C2E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3030003000_03.png with error HTTP Error 404: Not Found
ERROR:root:Bad Url http://gbf-wiki.com/attach2/696D67_333033303030393030305F30335F66756C6C2E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3030009000_03.png with error HTTP Error 404: Not Found
ERROR:root:Bad Url http://gbf-wiki.com/attach2/696D67_333033303031323030305F30335F66756C6C2E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3030012000_03.png with error HTTP Error 404: Not Found
ERROR:root:Bad Url http://gbf-wiki.com/attach2/696D67_333033303031343030305F30335F66756C6C2E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3030014000_03.png with error HTTP Error 404: Not Found
ERROR:root:Bad Url http://gbf-wiki.com/attach2/696D67_333033303031383030305F30335F66756C6C2E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3030018000_03.png with error HTTP Error 404: Not Found
ERROR:root:Bad Url http://gbf-wiki.com/attach2/696D67_333033303032303030305F30335F66756C6C2E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3030020000_03.png with error HTTP Error 404: Not Found
ERROR:root:Bad Url http://gbf-wiki.com/attach2/696D67_333033303032363030305F30335F66756C6C2E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3030026000_03.png with error HTTP Error 404: Not Found
ERROR:root:Bad Url http://gbf-wiki.com/attach2/696D67_333033303033333030305F30335F66756C6C2E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3030033000_03.png with error HTTP Error 404: Not Found
ERROR:root:Bad Url http://gbf-wiki.com/attach2/696D67_333033303033393030305F30335F66756C6C2E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3030039000_03.png with error HTTP Error 404: Not Found
ERROR:root:Bad Url http://gbf-wiki.com/attach2/696D67_333033303037383030305F30335F66756C6C2E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3030078000_03.png with error HTTP Error 404: Not Found
ERROR:root:Bad Url http://gbf-wiki.com/attach2/696D67_333033303039363030305F30335F66756C6C2E706E67.png at path C:\Users\Jongaros\WebstormProjects\motocal\charaimgs\3030096000_03.png with error HTTP Error 404: Not Found
```

kei-gbf:
ah sorry, I forgot about the transform.

- [ ] 404 http://gbf-wiki.com/index.php?plugin=attach&refer=img&openfile=3040007000_03.png
- [x] 200 http://gbf-wiki.com/index.php?plugin=attach&refer=img&openfile=3040007000_03_full.png

I do not know this yet, why needed the transform, etc.

OrkunKocyigit:
> 
> 
> ah sorry, I forgot about the transform.
> 
>     * [ ]  404 http://gbf-wiki.com/index.php?plugin=attach&refer=img&openfile=3040007000_03.png
> 
>     * [x]  200 http://gbf-wiki.com/index.php?plugin=attach&refer=img&openfile=3040007000_03_full.png
> 
> 
> I do not know this yet, why needed the transform, etc.

I am assuming wiki has no standart while formatting their file names. If we can't fix it in wiki side I can fix it with brute forcing all permutations.

scripts/download_image.py
```diff
@@ -37,73 +37,62 @@
...
-        if not all(x.casefold() in var_names for x in REQUIRED_REPORT_VALUES):
-            raise AttributeError(
-                'Reporter methods have to have count,'
-                ' total and message arguments')
 
 
-def _progress_reporter(count: int, total: int, message: FutureResult,
-                       bar_len: int = 45):
+def _progress_reporter(count: int, total: int, future_result: FutureResult,
+                       bar_len: int = 40, multiple_lines: bool = True):
```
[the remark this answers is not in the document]
OrkunKocyigit: I renamed future_result to result

scripts/download_image.py
```diff
@@ -37,73 +37,62 @@
...
-    print('[%4d/%4d] Download %s' % (count, total, message.url), flush=True)
+    >>> _plain_reporter(1, 100, FutureResult(r'url', r'path'))
+    [   1/ 100] Download url
+    """
+    if multiple_lines:
+        print('[%4d/%4d] Download %s' % (count, total, future_result.url),
+              flush=True)
+    else:
+        print('[%4d/%4d] Download %s' % (count, total, future_result.url),
+              flush=True, end='\r')
```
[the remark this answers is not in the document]
OrkunKocyigit: Created a new variable:
_report_printer = functools.partial(print, flush=True)
Changed method logic like this:
    if not multiple_lines:
        _report_printer.keywords['end'] = '\r'

scripts/download_image.py
```diff
@@ -136,6 +126,11 @@ def download_image(url: str, path: str, _retry_count: int = 3,
     :param _timeout: Connection timeout between client and server
     :param _wait_interval: Time to wait before retry
     :return: url and path back for reporting
+    >>> download_image(
+    ... 'http://gbf-wiki.com/attach2/696D67_313034303830393530302E706E67.png',
+    ... 'C://motocal//imgs//1040809500.png')
+    FutureResult(url='http://gbf-wiki.com/attach2/696D67_31303430383039353030\
+2E706E67.png', path='C://motocal//imgs//1040809500.png')
```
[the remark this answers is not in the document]
OrkunKocyigit: Changed

kei-gbf:
brute forcing ... I want to avoid that as possible.

```
> grep _03 haraImageWikiURLList.txt | wc
72
```
seem it's not all, need the fix url.
I do not know wiki's file naming rule.

This can be another issue, it's not by this PR's changes.

OrkunKocyigit:
> 
> 
> several minor clean up.
> 
> _read_lines can be read_lines now, its not module's top-level so not exposed.
> _max_workers as well.
> 
> ```diff
> - logging.error("Bad Url %s at path %s" % (url, path))
> + logging.error("Bad Url %s at path %s", url, path)
> ```
> 
> there are some using % operator in logging.error
> 
> ```
> Debug e.g: python download_image.py arm wiki tmp -d plain
> ```
> 
> can update, -c option?

This fails unit test.

Expected:
    [----------------------------------------] 1.0% ...                path
Got:
    [%s] %s%s ...%20s ---------------------------------------- 1.0 % path

kei-gbf:
> This fails unit test.

it was not about print(),  logging.error()

kei-gbf:
logging's reporter can take various parameters.

```python
logging.error("Bad Url %s at path %s with error %s" % (url, path, error))

logging.error("Bad Url %(url)s at %(path)s with error %(error)s", vars())

# this way is useful when need translate the message,
# and parameter order change by the locale.
```

kei-gbf:
can use real path, that give info what file to download.

```python
FutureResult(url='http://example.com/attach2/696D67_31303430383039353030\
2E706E67.png', path='C://motocal//imgs//1040809500.png')
```

I want to add `# doctest: +SKIP` here for skipping the test the external resource
the code will stay as sample document code in this case.

doctest directive official document does not explain the example though,
it's not undocumented feature, it used widely.
Can find in their test https://github.com/python/cpython/blob/master/Lib/test/test_doctest.py

kei-gbf:
logging.error(f"Bad Url {:url} at path {path} with error {error}")

is better, if python version is ok.

kei-gbf:
> so actual content is hosted url attach2/'img'_'filename'.png where img and filename is in HEX

thx information. I will try it mysel
later. give me a time.

kei-gbf:
> :param result: Message to print to console.

document is not true. it's sound like message is a text?
reporter function has the actual message in the message format,
the parameter is now just url and path.

scripts/download_image.py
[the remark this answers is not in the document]
kei-gbf: drop multiple_lines keyword argument, and
```python
if not multiple_lines:
    _report_printer.keywords['end'] = '\r'
```
no need this twice in each reporter functions.
can move this to main function?

now those reporter has an extra outer scope variable, "_print_reporter"
it make the function not portable, but it's ok if use with-in this script only.

kei-gbf:
- [ ] # doctest: +SKIP for check_image_url doctest
- [ ] update __doc__ comment: Debug e.g: python download_image.py arm wiki tmp -d plain
- [ ] multiple_lines option
- [x] logging.error
- [x] update README/README-ja (remove the -t argument)

scripts/download_image.py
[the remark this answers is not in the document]
kei-gbf: I noted in PR comment, # doctest: +SKIP and append path to example.com

kei-gbf:
> file_name = file_name.replace('_03.png', '_03_full.png')

I am not sure about this yet, now I can not test myself.
remove if this did not solve.

kei-gbf:
about this _report_printer, outer scope has store the state.

reporter is run in main thread, no problem to run the code
but for test it's not thread-safe code, dependency on outer scope's state.

```python
if not multiple_lines:
   _report_printer.keywords['end'] = '\r' 
```

e.g. (I know now we don't have this situation)
- worker-thread-A tests multi lines code
- worker-thread-B tests non multi lines code
if thread shares the keyword argument state,
the keyword arg may changed unexpected timing by another thread.

OrkunKocyigit:
Right now -d is left to dry-run, I can change name of check to debug but in this case dry-run have to be updated.

kei-gbf:
> Right now -d is left to dry-run, I can change name of check to debug but in this case dry-run have to be updated.

It's ok if not outdated, my memory did not update the latest version

OrkunKocyigit:
Print method itself is not thread safe. Ideally instead of using funct tools I can dump print function into a method like this:
_print_lock = Lock()
def _report_print(str, flush=True, multi=True):
    with _print_lock:
        print(str, flush=flush, end='\r' if not multi else '')

kei-gbf:
no need lock, this is a general coding topic
outer scope has state is like a global variable issue, makes test harder.

I think print is thread-safe in python3 (unless rewrite sys.stdout) but can't find the source to show you.
anyway, state-less is better if it's possible.

logging module is thread-safe for testing.

kei-gbf:
you can just remove the lock, we does not have this issue
until program become more large or many reporters implement.

keep a code simple compact, no unnecessary optimization.

state-less code just make easy to write test,
the state (ref in global) will be given as parameter,
and it ensure not effect to another thread.

kei-gbf:
is there situation _quiet_reporter take the multi argument?

----

about download from game, I worry if it was too much access.
mention in document or isn't it better prepare slow download option?
1 worker and wait interval each task.

scripts/download_image.py
```diff
@@ -2,10 +2,10 @@
 """
 Motocal Image download script.
 
-Usage: python download_image.py arm
-       python download_image.py chara
+Usage: python3 download_image.py arm
+       python3 download_image.py chara
 
-Debug e.g: python download_image.py arm wiki tmp -d plain
+Debug e.g: python3 download_image.py arm -c
```
kei-gbf: it's a simple good :+1:

OrkunKocyigit:
> 
> 
> is there situation _quiet_reporter take the multi argument?
> 
> about download from game, I worry if it was too much access.
> mention in document or isn't it better prepare slow download option?
> 1 worker and wait interval each task.
For internal debug I think multi will be always true.
I can print warning to console in case source is game and number of worker threads bigger than number x.

kei-gbf:
it almost finished big tasks? thank you a lot works.


https://deepsource.io/blog/8-new-python-antipatterns/

> 2. Having type information in a variable name
> Developers struggle with naming variables, of course. Sometimes it may happen that you add the type information in the variable name, which can be confusing for other developers, or even your future self, who can act on the variable assuming that it is of the type mentioned in its name. It is recommended to give meaningful names to variables suiting the context.

```python
url_list: list
```

there is case even anti patterns say, but sometime type info is still usefull.
however, we have annotation, it can be "urls" now. coz list twice.

----

reporter -> report
counter -> count

names can be verb as it's callable.

i was missing in list
functools.partial(next, itertools.count(start=1)) -> count = itertools.count(start=1) and next(count)

kei-gbf:
> I can print warning to console in case source is game and number of worker threads bigger than number x.

:+1:

kei-gbf:
'img'.encode('utf-8').hex().upper() is always same value.
encoded letters can be in template. and leave a note in comment it's from {img}_{file_name}

```
>>> b'img'.hex().upper()
'696D67'
```

```python
url = r'http://gbf-wiki.com/attach2/696D67_{}.png'
file_name = file_name.replace('_03.png', '_03_full.png') # was this ok? if not, please drop for temporary.
return url.format(file_name.encode('utf-8').hex().upper())
```

----

multi does not work well now? sorry I did not run while review, I don't know which changes this.
I got a time now.

py scripts/download_image.py arm

キャプチャ3

scripts/download_image.py
```diff
@@ -187,6 +188,29 @@ def main(argv: list):
     parser = _create_parser()
     options = parser.parse_args(argv)
 
+    if options.site == 'game' and options.workers > GAME_WORKER_LIMIT:
+        if options.y:
+            logging.info("Passing warning message...")
+        else:
```
kei-gbf: - single name 'y' can be long name
  - choice y/n is ok there, users does not want input long.

this section can be another function? because now it's a long else: block

```python
if yes:
    logging.info("Passing warning message...")
    return
choid = input(
```

OrkunKocyigit:
> 
> 
> 'img'.encode('utf-8').hex().upper() is always same value.
> encoded letters can be in template. and leave a note in comment it's from {img}_{file_name}
> 
> ```
> >>> b'img'.hex().upper()
> '696D67'
> ```
> 
> ```python
> url = r'http://gbf-wiki.com/attach2/696D67_{}.png'
> file_name = file_name.replace('_03.png', '_03_full.png') # was this ok? if not, please drop for temporary.
> return url.format(file_name.encode('utf-8').hex().upper())
> ```
> 
> multi does not work well now? sorry I did not run while review, I don't know which changes this.
> I got a time now.
> 
> py scripts/download_image.py arm
> 
> キャプチャ3

fixed

scripts/download_image.py
```diff
@@ -254,10 +278,10 @@ def scan_file_for_download_list(url_list: list, output: str, force: bool):
 
     with ThreadPoolExecutor(max_workers=max_workers) as executor:
         submit = functools.partial(executor.submit, worker_method)
-        counter = functools.partial(next, itertools.count(start=1))
-        reporter = REPORT_TYPE.get(options.reporter)
+        count = itertools.count(start=1)
+        report = REPORT_TYPE.get(options.reporter)
         for future in as_completed(itertools.starmap(submit, items)):
-            reporter(counter(), total, future.result())
+            report(next(count), total, future.result())
```
kei-gbf: :+1:

kei-gbf:
thx, can you pls edit the comment to remove screenshot image in quote again.

- [x] multiple_lines option (note in review comment)
- [x] update document  :param result: Message to print to console.
- [x] drop functools.partial for itertools.count (EDITED) 
- [x] warnings for download from game
  - [ ] separate the function 

----
REPORT_TYPE = { and 
TXT_SOURCE had different indent style.
in Python, I like REPORT_TYPE style with 4spaces indent. (now it's 8 or tab?)
REPORT_TYPE style make length longer, long variable name make the long unnecessary indent spaces.

SAVE_DIR is ok short in single line.

scripts/download_image.py
OrkunKocyigit: Changed main method to this:
```
    if options.site == 'game' and options.workers > GAME_WORKER_LIMIT:
        if options.y:
            logging.info("Passing warning message...")
        elif show_game_download_warning(options.workers):
            return
```
And show_game_download_warning is:
```
def show_game_download_warning(workers: int,
                               input_fn: callable = input) -> bool:
    """
    Shows warning in case user selected game as a source
    :param input_fn: Input function for test or file reading
    :param workers: Number of workers
    :return:
    >>> show_game_download_warning(10, input_fn=lambda x: "y")
    False
    >>> show_game_download_warning(10, input_fn=lambda x: "n")
    True
    """
    choice = input_fn(
            'WARNING: You specified {} workers which is over limit {'
            '} to '
            'download images from game website. This may lead you to '
            'get '
            'banned. If you accept it enter Y or enter N to exit '
            'script. '
            'If not please define number of threads with -w '
            'parameter. '
            'If you want to skip this warning you can enter -y while '
            'calling script.:'.format(workers,
                                      GAME_WORKER_LIMIT)).lower()
    while choice not in ['y', 'n']:
        choice = input_fn('Either enter Y or N as an argument:').lower()
    if choice == 'n':
        return True
    else:
        logging.info('Warning accepted, script will continue')
        return False
```

kei-gbf:
pycodestyle (pep8) check is ok 
got total 916, error is Lily? (this is chara link issue)

- warning message can contain line break? and space after prompt ":" 
- it input "-y" not "y" ?

scripts/download_image.py
```diff
@@ -102,11 +102,8 @@ def _do_nothing(*args, **kwargs):
     pass
 
 
-REPORT_TYPE = {
-        'progress': _progress_reporter,
-        'plain': _plain_reporter,
-        'quiet': _quiet_reporter,
-}
+REPORT_TYPE = {'progress': _progress_reporter, 'plain': _plain_reporter,
+               'quiet': _quiet_reporter}
```
kei-gbf: no that make line length long

```python
REPORT_TYPE = {
    'progress': _progress_reporter,
    'plain': _plain_reporter,
     'quiet': _quiet_reporter,
}
```

kei-gbf: number of indent for REPORT_TYPE
and change request for TXT_SOURCE

scripts/download_image.py
kei-gbf: NO return True? how about caller change 
if not accept_game_download_warning(): ...

scripts/download_image.py
```diff
@@ -285,6 +265,39 @@ def scan_file_for_download_list(urls: list, output: str, force: bool):
...
+            '} to '
+            'download images from game website. This may lead you to '
+            'get '
+            'banned. If you accept it enter Y or enter N to exit '
+            'script. '
+            'If not please define number of threads with -w '
+            'parameter. '
+            'If you want to skip this warning you can enter -y while '
+            'calling script.:'.format(workers,
+                                      GAME_WORKER_LIMIT)).lower()
```
kei-gbf: 'get ', '} to ' is too small chunk.
If line length was the limit, this text can move to global (top-level)
it's not became state, Python's string is an immutable object.

OrkunKocyigit: Fixed by moving string into variable.

scripts/download_image.py
OrkunKocyigit: Fixed
```
REPORT_TYPE = {
        'progress': _progress_reporter,
        'plain': _plain_reporter,
        'quiet': _quiet_reporter,
}

TXT_SOURCE = {
        'arm-wiki': "armImageWikiURLList.txt",
        'chara-wiki': "charaImageWikiURLList.txt",
        'arm-game': "armImageGameURLList.txt",
        'chara-game': "charaImageGameURLList.txt"
}
```

scripts/download_image.py
OrkunKocyigit:
```
    warning = 'WARNING: You specified {} workers which is over limit {} to ' \
              'download images from game website This may lead you to get ' \
              'banned. If you accept it enter Y or enter N to exit script. ' \
              'If not please define number of threads with -w parameter. If ' \
              'you want to skip this warning you can enter -y while calling ' \
              'script : '.format(workers, GAME_WORKER_LIMIT)
    warning = '\n'.join(warning[i:i + 92] for i in range(0, len(warning), 92))
```

scripts/download_image.py
```diff
@@ -277,18 +281,15 @@ def show_game_download_warning(workers: int,
...
-            'calling script.:'.format(workers,
-                                      GAME_WORKER_LIMIT)).lower()
+    warning = 'WARNING: You specified {} workers which is over limit {} to ' \
+              'download images from game website This may lead you to get ' \
+              'banned. If you accept it enter Y or enter N to exit script. ' \
+              'If not please define number of threads with -w parameter. If ' \
+              'you want to skip this warning you can enter -y while calling ' \
+              'script : '.format(
+        workers, GAME_WORKER_LIMIT)
+    warning = '\n'.join(warning[i:i + 92] for i in range(0, len(warning), 92))
```
kei-gbf: no, don't want this trick. write in top-level 
```python
WARNING_MESSAGE = """\
    WARNING: 
     ...
"""
```

textwrap.dedent(text) helps this.

kei-gbf:
- [ ] return True for N ? https://github.com/MotocalDevelopers/motocal/pull/185#discussion_r283142808
- [ ] warning message https://github.com/MotocalDevelopers/motocal/pull/185#discussion_r283143442
  - I mention textwrap library

kei-gbf:
REPORT_TYPE/TXT_SOURC  indent is 8 ? or tab? should be width 4 spaces

kei-gbf:
scan_file_for_download_list function now no options dependency,
can go to top-level?

'wiki' in url
given option.site then site == 'wiki' was ok.
any url witch contains wiki will be transform, it not happen though.
unit test may pass "game-server/file-name-wiki.png" case.

python use intern (sys.intern)
that can check only object id, no text match for const strings.
(cos python string is immutable)

scripts/download_image.py
OrkunKocyigit: Fixed

scripts/download_image.py
OrkunKocyigit: Changed with auto terminal size splitting:
```
    choice = input_fn('\n'.join(textwrap.wrap(
        textwrap.dedent(WARNING_MESSAGE.format(workers, GAME_WORKER_LIMIT)),
        shutil.get_terminal_size().columns))).lower()
```
WARNING: You specified 10 workers which is over limit 1 to download images from
game website This may lead you to get banned. If you accept it enter Y or enter
N to exit script. If not please define number of threads with -w parameter. If
you want to skip this warning you can enter -y while calling script :

kei-gbf:
@OrkunKocyigit 

ah a bit late
I was going to make PR, https://github.com/kei-gbf/motocal/tree/review-pr185

OrkunKocyigit:
It's okay, I'll do manual merge for parts.

kei-gbf:
can you pick up some if you feel good.

except:
get_args = operator.attrgetter("output", "force", "site")

was save lines for expand variables.
used in short script when don't want to type many same prefix "options."
I used coz with-open scope is few lines. should not abuse this kind of code.
you can drop this, because this code loose type information.

kei-gbf:
- [ ] shutil.get_terminal_size may fail, can drop this
- [ ] two .lower() -> while choice.lower() once
- [ ] tidy WARNING_MESSAGE

scripts/download_image.py
kei-gbf: can you remove shutil.get_terminal_size() and tidy WARNING_MESSAGE
readable in source code too.

kei-gbf: an no, choice == 'y' may False

kei-gbf:
sorry it was my mistake, about .lower()

scripts/download_image.py
```diff
@@ -295,13 +295,10 @@ def accept_game_download_warning(workers: int,
...
     """
     import textwrap
-    import shutil
-    choice = input_fn('\n'.join(textwrap.wrap(
-            textwrap.dedent(
-                    WARNING_MESSAGE.format(workers, GAME_WORKER_LIMIT)),
-            shutil.get_terminal_size().columns))).lower()
-    while choice not in ['y', 'n']:
-        choice = input_fn('Either enter Y or N as an argument:').lower()
+    choice = input_fn('\n'.join(textwrap.wrap(textwrap.dedent(
```
kei-gbf: '\n'.join(textwrap.wrap(   need this?
we can keep the WARING_MESSAGE text length short.

workers and limit are only number, length will not be too long
unless user input huge number. btw range check for the options?

scripts/download_image.py
```diff
@@ -295,13 +295,10 @@ def accept_game_download_warning(workers: int,
...
+        WARNING_MESSAGE.format(workers, GAME_WORKER_LIMIT)))))
+    while choice.lower() not in ['y', 'n']:
+        choice = input_fn('Either enter Y or N as an argument:')
     return choice == 'y'
```
kei-gbf: can you revert here, my bug (choice may be 'Y') upper letter
or return choice in "Yy" ... strange though.  
or return choice in {'Y', 'y'}  # set-iteral

scripts/download_image.py
OrkunKocyigit: Can you repull again?

scripts/download_image.py
OrkunKocyigit: Fixed return choice.lower() == 'y'

kei-gbf:
can you check max workers number ?
it's less than total before run ThreadPoolExecutor,
but warning show large number user input.

kei-gbf:
> Can you repull again?

sorry, I did not notice this, still need ?

OrkunKocyigit:
> 
> 
> can you check max workers number ?
> it's less than total before run ThreadPoolExecutor,
> but warning show large number user input.

I cannot recreate this. By default script creates 10 workers,GAME_WORKER_LIMIT set to 1.

kei-gbf:
it's ok for default case. just output text layout.

download_image.py arm -s game -w 100000000000000000000
will show over line length in the warning message.
increase the line length by user input, but not much problem.

kei-gbf:
it's ok for default case. just output text layout.

download_image.py arm -s game -w 100000000000000000000
will show over line length in the warning message.
increase the line length by user input, but not much problem.

OrkunKocyigit:
> 
> 
> it's ok for default case. just output text layout.
> 
> download_image.py arm -s game -w 100000000000000000000
> will show over line length in the warning message.
> increase the line length by user input, but not much problem.

I am pretty sure trying to create 100000000000000000000 threads will crash/lock basically any computer

kei-gbf:
max_workers = max(1, min(total, options.workers))

will protect the case, the total will not be such big number.
of coz it's good if check the max value explicitly in options section.

kei-gbf:
[ANN] for Japanese
画像ダウンロード スクリプトの並行処理対応、レビュー完了です。

一枚だけ正常に拾えない画像がありますが、このPRの影響ではないので別issueにします。
早すぎてアクセス過剰で 503 エラーが返ることもあったので、
ゲームサーバへアクセスする場合は、事前に警告がでてスレッド数も制限されるようになりました。

後、長くなったけど殆どの話題は、コードの整理など。

kei-gbf:
@OrkunKocyigit 
I'll finish review if you don't have extra tasks,
thx long conversation, the code quality is now pretty good.
(lack of tests but too much enough as a single script)

motocal javascript code also should have documents/tests,
I feel regression test for each PR review is now costs. I want automate it.

OrkunKocyigit:
Yep it is good enough for it is purpose.

PR #200, opened 2019-05-15
Added ougi damage bonus buff to characters

OrkunKocyigit:
Currently it is possible to add ougi bonus to none other than main character. With this patch I aim to add this functionality to characters as well for EMP bonuses.


kei-gbf:
Hello, I test it works good.

- [x] save data compatibility
  - [x] TEST PASS: I output JSON and remove "ougiDamageChara" line, then load back ... ok
         no errors on develop console.
- [ ] ougiDamageBuff: charaBuffList["ougiDamageChara"]
  - [ ] how about charaBuffList key also "ougiDamageBuff" ?
         other buff's keys also has "Buff" suffix.
         I'm not sure this naming rules, other has "chara" prefix.
         Pros: at least, grep /ougiDamageBuff/ can show related lines.
- [ ] (not change request), just research about ougiDamage up fields
       There are Buff and LB(EMP) category, Djeeta 's buff does not have ougi damage
       Djeeta has a party buff too, missing ougi damage.
   maybe next or another topic?

OrkunKocyigit:
Updated key name. I was thinking eventually adding extended master ring segment to reflect that table
image
https://docs.google.com/spreadsheets/d/1uehv6CTXeepJjUVOPL03FywwMVlT1QW62MZO-R21ndU/edit#gid=1519281220

From what I can find rings provide bonuses additionally to extended mastery perks instead of running different multiplier.

kei-gbf:
Can you consider this,
I worry current fields implementation increase JSON data size.
If the approach was like this:

ringAtk: 300,
ringHP: 150, 
: 
all characters will have unnecessary fields if they does not equip ring.

- e.g. ring: { atk: 300, hp: 150, ougiDamage: 10, dadge: 1}
- if character has no ring then `ring: {}`  can save JSON data size.

There are saved data on server, we can't break the compatibility.
but new fields can improve this.

----
I'll be busy for guild war next week, lets enjoy!

kei-gbf:
I got NaN when no weapons, add character then set ougiDamageBuff

src/global_logic.js
```diff
@@ -1840,6 +1840,7 @@ module.exports.getInitialTotals = function (prof, chara, summon) {
                 daBuff: 0.0,
                 taBuff: 0.0,
                 ougiGageBuff: 0.0,
+                ougiDamageChara: 0.0,
```
kei-gbf: Chara -> Buff here
this make missing initialize then got NaN

OrkunKocyigit: Oops missed that one. Updated

OrkunKocyigit:
> Can you consider this,
> I worry current fields implementation increase JSON data size.
> If the approach was like this:
> 
> ringAtk: 300,
> ringHP: 150,
> :
> all characters will have unnecessary fields if they does not equip ring.
> 
> * e.g. ring: { atk: 300, hp: 150, ougiDamage: 10, dadge: 1}
> * if character has no ring then `ring: {}`  can save JSON data size.
> 
> There are saved data on server, we can't break the compatibility.
> but new fields can improve this.
> 
> I'll be busy for guild war next week, lets enjoy!

I have not considered that but I think that would be swell. Good luck with GW ^^

PR #222, opened 2019-06-01
Added dark opus weapon limit (1 per grid)

OrkunKocyigit:


kei-gbf:
@OrkunKocyigit 
This is not a request, just question and extra topic.

Can you do the same for Hollowsky weapons ?
I do not see this Motocal code at all, Cosmos is now a grid can have multi cosmos weapons.

I will review/test later, just replay and question for now.

src/global_logic.js
```diff
@@ -33,6 +33,11 @@ module.exports.isCosmos = function (arm) {
         (skilltypes[arm.skill3] != undefined && skilltypes[arm.skill3].type == "cosmosArm");
 };
 
+module.exports.isDarkOpus = function (arm) {
+    return (arm.skill1 != undefined && arm.skill1 == "normalKamuiL") ||
+        (arm.skill1 != undefined && arm.skill1 == "magnaKamuiL");
+};
+
```
kei-gbf: `arm.skill1 != undefined` double checking
- [ ] `arm.skill1 != undefined && arm.skill1.endsWith("KamuiL")`
  - but "exKamuiL" may hit this, regex for more strict check.
- [ ] `arm.skill1 != undefined && arm.skill1.match(/^(normal|magna)KamuiL$/);`
  - generate match object unnecessary
- [x] `arm.skill1 != undefined && ((arm.skill1 == "normalKamuiL") || (arm.skill1 == "magnaKamuiL"))`
  - simple is good

Motocal Weapons form allow users to change the skill.
- Users can set Kamui(L) to another weapons.
- in future, if another kamui(L) weapon was added.

Current isDarkOpus work for default DarkOpus weapon,
but no support custom weapons. This can be future topic, new weapon may break this.

OrkunKocyigit:
I actually considered that but only fool proof thing I can find is just checking by id. I can do that if that is okay.

kei-gbf:
I could not find id in armData.json. what id did you think? ImageURL?

- weapon name is not unique, localized and user can edit
- imageURL ... I saw in gbf.wiki use image filename as ID.

Now I read code, other weapons cosmos use isCosmos() checked skills
it was ok unique skill for Cosmos.
For DarkOpus Kamui is like a general skill, other weapons may have.

Point I worry here.
for users, treat kamuiL as DarkOpus weapon is an undocumented.

Just idea for another approach
Set something blank skill to skill2,skill3 (Ultima weapon does not have skill3)
that only DarkOpus can have. is it possible?

OrkunKocyigit:
These are all I can check:
image
Only think that looks usable is name.

src/global_logic.js
OrkunKocyigit: I added id based comparison on my last commit. That should fix all problems.
New method:
module.exports.isDarkOpus = function (arm) {
    return darkOpusWeapons.some(function (value) {
        return arm.id == value;
    });
};
Dark opus weapons:
var darkOpusWeapons = [
    "1040017000",
    "1040017100",
    "1040212500",
    "1040212600",
    "1040310600",
    "1040310700",
    "1040415000",
    "1040415100",
    "1040809400",
    "1040809500",
    "1040911000",
    "1040911100"
];

ID is parsed from image_url with regex.

OrkunKocyigit:
> 
> 
> @OrkunKocyigit
> This is not a request, just question and extra topic.
> 
> Can you do the same for Hollowsky weapons ?
> I do not see this Motocal code at all, Cosmos is now a grid can have multi cosmos weapons.
> 
> I will review/test later, just replay and question for now.

I can easily do that if you can verify the requirements. From what I understand:

- [ ] Only one type of hollowsky weapon can be in grid in any given time.
- [ ] Only one hollowsky weapons of each type.
- [ ] Only one type of cosmic weapon can be in grid in any given time.
- [ ] Only one cosmic weapons of each type.

I think those are true for Qilin and Huanlong and Rose Petal weapons too if I am not mistaken.

kei-gbf:
- [ ] reduce calculation
  treatAddArmFromTemplate can set isDarkOpus flag but id may be used for other.
  I am not sure how often isDarkOpus is called. I will research this in my review.

> Only one type of hollowsky weapon can be in grid in any given time.
> Only one hollowsky weapons of each type.

I am not sure this, the "grid" is the row in Motocal result? or game grid.

here, I assume use-case scenario:
in Motocal, Users can add different types of hollowsky in the list,
the result has only one type in each row.

I had discuss this topic with @Groovman before. in another PR
There is another opinion too, users may want to calc that assume unusual situation.
like Motocal allows change weapon's skill and calculate IF the weapon had another skill etc.

> Only one type of cosmic weapon can be in grid in any given time.
> Only one cosmic weapons of each type.

this is about the game system, 
- players could not get more than one cosmos weapon before.
- it's changed by update. now it's ok just skill included check.


> I think those are true for Qilin and Huanlong and Rose Petal weapons too if I am not mistaken.

Yes they are also unique weapon. but Cosmos(cosmic?) weapons are now different.
it's possible to set (useless) grid. I know it's rare case, no one want to calc this in Motocal
so I think it's low priority stuff.

cosmic

src/armlist.js
```diff
@@ -348,6 +349,12 @@ var Arm = CreateClass({
         this.props.onChange(this.props.id, state, false);
     },
     treatAddArmFromTemplate: function (state, newarm, considerNum) {
+        var pattern = new RegExp("(\\d+)","i");
+        var matches = pattern.exec(newarm.imageURL);
```
kei-gbf: `var matches = newarm.imageURL.match(/(\d+)/);`

no need ignore case option for number.

src/global_logic.js
```diff
@@ -34,8 +35,9 @@ module.exports.isCosmos = function (arm) {
 };
 
 module.exports.isDarkOpus = function (arm) {
-    return (arm.skill1 != undefined && arm.skill1 == "normalKamuiL") ||
-        (arm.skill1 != undefined && arm.skill1 == "magnaKamuiL");
+    return darkOpusWeapons.some(function (value) {
+        return arm.id == value;
+    });
```
kei-gbf: I am not sure this, if Babel allow more simple way
- darkOpusWeapons.some(value => arm.id == value);
- darkOpusWeapons.includes(value)

kei-gbf:
@OrkunKocyigit 
Hollowsky and Cosmos weapons are extra topics,
I will review Dark Opus in this PR.

----
in ID(imageURL) check approach, 

- cost of how often isDarkOpus() is called 
- stored in state["id"] then save data has the data?
- old save data which had dark opus weapon before, are out of this support?
  if this make users another confuse, skill check approach is better.
  I don't know how many save data which has dark opus weapon created.

----
in skill1 kamuiL check approach

```js
    return (arm.skill1 != undefined && arm.skill1 == "normalKamuiL") ||
           (arm.skill1 != undefined && arm.skill1 == "magnaKamuiL");
```

- `arm.skill1 != undefined` check is duplicated
- skill check approach is ok for now, (until new kamuiL weapon added)
  - except users may feel it's undocumented.

OrkunKocyigit:
@kei-gbf 
Well I can remove middle man with:
- [ ] Remove id attribute and isDarkOpus methods
- [ ] Put darkOpus attribute flag in arm
Would that be fine?

kei-gbf:
@OrkunKocyigit 

about the calc cost, yes.

but the next issue is save data compatibility since it's stored in state.
I'm not sure how much we should care this,
at this point, I feel kamuiL skill check was ok.

the 3rd approach i shown above is set skill2,3
that something opus-non, treat it as isDarkOpus flag
when skill2,3 was set, skill3 check opusKey can be the flag.

darkopus


----
about Hollowsky weapon exclusion
#132 (long PR but just the beginning part)

Groovman:
Easiest solution check the arms name prefix/suffix to find the opus weapon and add to the magna/normalKamuie section a check if a opus weapon was added and check a bitmask like it is done for ultima and hollowsky already. with magna/normal + the element, and just or the whole bitmask to check if you can add one.

kei-gbf:
isDarkOpus
- [ ] skill1 must be normal of magna KamuiL
- [ ] name contains that opus weapon
  - [ ] must be locale safe check which pass "ja" and "en" cases

implement the name check at this moment?
unlike store in state approach, this does not have a side effect.
users can edit the name is ... just became known issue.

OrkunKocyigit:
Changed logic to this:
```
module.exports.isDarkOpus = function (arm) {
    return (GlobalConst.opusNames.some(value => arm.name.includes(value)) && skilltypes[arm.skill1]) &&
        ["magnaKamuiL","normalKamuiL"].some(value => arm.skill1 == value);
};
```

kei-gbf:
- [x] GlobalConst.opusNames.some(value => arm.name.includes(value))
  - looks good use of some()
- [ ] skilltypes[arm.skill1])
  - need this check?
- [ ] ["magnaKamuiL","normalKamuiL"].some(value => arm.skill1 == value);
  - how about includes?

kei-gbf:
I setup the playground for the isDarkOpus function and unit tests
https://codepen.io/anon/pen/xNMKLp?editors=0010

Anyone can "Fork" and "Edit"

OrkunKocyigit:
@kei-gbf 
>       * looks good use of some() and includes()
>       * [ ]  no need to check arm.name undefined case?
I tested it and undefined case always returns false. I thought that was acceptable solution.
>     * [ ]  skilltypes[arm.skill1])
>       * need this check?
My code doesn't have that change. I am guessing it is about Cosmos weapons. Cosmos weapon skills are all named Cosmos "Weapon Type" so I am guessing they all share same name. For dark opus that is not the case.
>     * [ ]  ["magnaKamuiL","normalKamuiL"].some(value => arm.skill1 == value);
>       
>       * how about includes?
>     * ~calc order, check skill1 may be a bit fast, if isDarkOpus is called many times.~
> 
>     * ~kamui array, JavaScript create the array object run-time, it can be outside of function as const.~
> 
> 
> EDIT: Most of case when name check passed, then skill1 check also ok,
> it's not make the temporary array so many times in this order.
> and this is micro optimization which can be unnecessary, it's because I don't know how often
> the function is called yet.
This function has two uses cases:
- Function is called n times when weapon list changes to see if combinations need to be recalculate. 
- Function is called n times when recalculating all permutations.
In both use cases n is equal to number of different weapons in armlist and limited to 20.

I agree name check is probably enough. I'll remove skill check.

OrkunKocyigit:
> 
> 
> I setup the playground for the isDarkOpus function and unit tests
> https://codepen.io/anon/pen/xNMKLp?editors=0010
> 
> Anyone can "Fork" and "Edit"

Thanks. I added your test cases to global_logic.test.js

kei-gbf:
Thanks info, give me a time i  need to understand that how often the func is called.
and Question about the Timing
when user change the weapon name ... it's still dark opus ? or re-calculate the check?
(change weapon name was not effect to result usually)

----
Review for tests, a bit more:

- no need opusName import
  - JavaScript has a closure, it's safe to remove.
- test('should be dark opus weapon' ...
  - Can you add more tests, four cases for each opusNames, but without import opusNames.
    This is a test for opusNames, so not do opusNames.each.
- skill1 check is ok, or redundant
  - if no skill check ... then test data also can drop skill field
  - I just show this as info. it's ok to drop skill check
    Test has a side it's write a spec, so if test checked skill 
    Readers will see all dark opus weapon has skill1 is kamuiL
    I think it's less case skill is checked, but still ok to stay there as spec.
  - Merit of drop skill1 check, users can calc the case IF dark opus weapon had another skill1
- sampleArmJA, sampleArmEN is assume it's real data. (pasted in codepen was for demo)
  - it should load data from armData.json In future, when we setup E2E tests.
  - now in Unit test, we can drop the two. that test does not check much than name
    E2E test (which load from armData.json) will check armData.json is not broken.
    This is ok skip for now in this PR, E2E tests can be another topic.
- arm.name is undefined is rare case, as I can think is only
   when users modify name field in save data and then.
   it's safe to check even if no need, that may protect an unexpected situation.

OrkunKocyigit:
Removed imports and added a lot of troll use cases in both languages

kei-gbf:
@OrkunKocyigit 

> opusNames
I missed to tell that context, it was for test. import in global_logic was ok

mixing all arm data is ...
that lose the group information in the test result report.
I suggest at least 3 cases.

- [ ] test should be dark opus
     - test all dark opus names can pass
- [ ] test other weapons should not be dark opus
     - legal another weapon which is not dark opus
- [ ] test broken weapon data
     - this test cases check illegal, exception cases like undefined field etc

OrkunKocyigit:
@kei-gbf I updated categories as you requested.

kei-gbf:
darkopus

report display `[object Object]`
I took this in previous commit.


```jsx
const darkOputNames = [
    "Sword of Renunciation",
    :
]
```
and pass `name` to test case `expect(isDarkOpus({name: name}).toBeTruthy();`

----
I noticed that mistake test.each in Haisui/Konshin, I will fix that in other PR.
test.each made the report long. it was a demerit side.

OrkunKocyigit:
@kei-gbf 
Weird doesn't happen with my setup, I'll check and figure it out.
image

I assume you wanna see only 3 checkboxes instead of each data row right?

kei-gbf:
categories :+1:
now no need expected parameter,
you can use toBeTruthy() and toBeFalsy() each category

I think invalidArms test can be single report,
no need to know the details in report.

```jsx
test('Checking invalid arms', () => {
    expect(isDarkOpus({})).toBeFalsy();
    : snip
});
```

kei-gbf:
> I assume you wanna see only 3 checkboxes instead of each data row right?

Yes, I misunderstood about `test.each` at this point.
Jest framework may have option for sub group ... I am not sure this for now.

> display [object Object]

oh, maybe my environment problem. node/libraries version or terminal.

OrkunKocyigit:
Here is latest log from Travis.
image

kei-gbf:
report result is nice but not for-loop in the test, 
that has known issue, when test failed error info became useless.

Python unit-test had "subTest" context manager for that solution.
I search it now if Jest may have ...

kei-gbf:
test-fail

report can show recieved: true ... but don't know what parameter for isDarkOpus

so list all code is better in a test.

```jsx
test('Checking invalid arms', () => {
    expect(isDarkOpus()).toBeFalsy();
    expect(isDarkOpus({})).toBeFalsy();
    expect(isDarkOpus({name:""})).toBeFalsy();
};
```

kei-gbf:
@OrkunKocyigit 

- [ ] no need to import opusNames, it's unused in test file
  - global_logic.js could import that.
- [ ] const nonDarkOpusArms is now unused

I will review tomorrow. sleep time now

OrkunKocyigit:
Cool, thanks for your work

PR #223, opened 2019-06-01
Added Wedding Rings to Characters

OrkunKocyigit:
I am not sure about Japanese word for it so send my branch a PR if it is not good ^^


src/global_logic.js
```diff
@@ -446,6 +449,9 @@ module.exports.calcBasedOneSummon = function (summonind, prof, buff, totals) {
         if (key == "Djeeta") {
             hpCoeff += 0.01 * totals["Djeeta"]["job"].shugoBonus;
         }
+        if (totals[key]["EXLB"]["WED"]){
+            hpCoeff += 0.01 * 10;
```
kei-gbf: can be `hpCoeff += 0.10;` like other lines. 0.10 for 10% instead of 0.1

kei-gbf:
- [x] Extended Mastery Star Cap +10
  - [ ] (Question) currently, no other buff which  add to this field?
    - [x] AFAIK, Shield of Eternal Splendor is not implemented
- [x] ATK +10%
- [x] HP +10%
- [x] DMG Cap +5%
  - [x] ougiDamageLimit +5%
  - [x] damageLimit +5%

----
coding style
- [ ] missing semi-colons except HP.
- [ ] getCharaEXLB() space after keyword "if" and "else"
    https://eslint.org/docs/rules/keyword-spacing

src/global_logic.js
```diff
@@ -614,7 +614,7 @@ module.exports.calcBasedOneSummon = function (summonind, prof, buff, totals) {
         damageLimit += Math.min(0.20, totals[key]["normalDamageLimit"]);
         damageLimit += 0.01 * totalSummon["damageLimit"];
         if (totals[key]["EXLB"]["WED"]){
-            damageLimit += 0.05
+            damageLimit += 0.05;
```

src/chara.js
```diff
@@ -393,10 +393,8 @@ var Chara = CreateClass({
     handleSelectEvent: function (key, e) {
         var newState = this.state;
 
-        if (key == "isConsideredInAverage") {
-            newState[key] = (newState[key] == false) ? true : false
-        } else if (key == "EXLBWED") {
-            newState[key] = (newState[key] == false) ? true : false
+        if (key == "isConsideredInAverage" || key == "EXLBWED") {
+            newState[key] = (newState[key] == false)
```
kei-gbf: I do not know what do this, just javascript expression
- boolean context can be `!newState[key]` ?
  - although when newState[key] was 0, `!0` also true. then if condition can check the type
    - if (typeof newState[key] === 'boolean') { // key: isConsideredInAverage, EXLBWED cases
- switch (key) is another choice. if another type of key is add in future.

```jsx
switch (key) {
    case "isConsideredInAverage":
    case "EXLB":
        newState[key] = (newState[key] == false);
        break;
    default:
        newState[key] = e.target.value;
        break;
}
```

for now, it's only two keys then the current code is ok.
- [ ] just remember **";" at the end.**

kei-gbf: BTW, why inverse flag here?

kei-gbf:
LGTM

EXLBWED: false


EXLBWED: true


----
NOTE: the display ATK contains skill at this moment.

- reported on #214 
- added task to TODO list #151 
  - グラブルの表示攻撃力に合わせた普通の「攻撃力」と
    スキル込の「基礎ダメージ」で表示項目を分ける
  - en: Separate ATK fit to "Chara ATK" in game, and "Base damage" which contains skill effect.
    now Motocal display is the base damage. ATK +10% does not much to the display ATK

----
Known issue about initialize at loading old save data
I reported on #225, another fields recently added also has this issue.
won't fix in this PR. another PR will resolve this.

src/chara.js
kei-gbf: if e.target was Checkbox and e.target.checked,

switch logic per e.target then
when another key is add, no need to change here.

kagekiyo7:
> (Question) currently, no other buff which add to this field?

効果の内の別枠バフは大事なもの「十天光輝の盾」枠(統べ称号で貰えるやつ)と同じ枠ですね
https://docs.google.com/spreadsheets/d/1JHwDTzStEm2lHgXDGeoFQLHz3Q85oRyVlIGI2i7zXKo/edit#gid=1247958484

なので、できれば新しく別枠の変数を作ったほうがいいかと

src/global_logic.js
OrkunKocyigit: Updated on latest branch

src/chara.js
OrkunKocyigit: Updated to check if it is checkbox instead of keys on 1185c63

src/global_logic.js
```diff
@@ -1837,10 +1836,9 @@ function getCharaEXLB(chara) {
     Object.keys(EXLB).map((key) => {
         var exactKey = "EXLB" + key;
         if (exactKey in chara) {
-            if(exactKey == "EXLBWED"){
+            if(exactKey == "EXLBWED") {
```
kei-gbf: - [ ] one more space > "if ("
- [ ] other key check
  - all exactKey has "EXLB" prefix, then can be key == "WED" (or key === "WED")
  - or typedef chara[exactKey] === 'number' for parseInt

src/chara.js
```diff
@@ -394,7 +394,7 @@ var Chara = CreateClass({
     handleSelectEvent: function (key, e) {
         var newState = this.state;
 
-        if (key == "isConsideredInAverage" || key == "EXLBWED") {
+        if (e.target.type === "checkbox") {
```
kei-gbf: :+1:

`newState[key] = e.target.checked;` also worked

src/global_logic.js
kei-gbf: Object.keys(EXLB).forEach is better here, map() function creates an array, 
the returned value is unused in this code.

This is not change request for PR.
because, it's not only this function, I just note for now.

OrkunKocyigit: Fixed in ddda901

kei-gbf:
no need "EXLB" prefix for `key`, it `exactKey` had.

src/global_logic.js
OrkunKocyigit: Removed EXLB check in 1594a8f

src/chara.js
OrkunKocyigit: Changed in 1594a8f

src/global_logic.js
```diff
@@ -1836,7 +1836,7 @@ function getCharaEXLB(chara) {
     Object.keys(EXLB).map((key) => {
         var exactKey = "EXLB" + key;
         if (exactKey in chara) {
-            if (key === "EXLBWED") {
+            if (key === "WED") {
```

h-yasha:
just a note:
`Magic Torrent's 20% ATK Up is on the same modifier as the Wonder Shield of Eternal Splendor, but is a unique modifier for any other buff.` - https://gbf.wiki/Fif
and, `Gates of Demons: has an ATK Up modifier that's the same as the Wonder Shield of Eternal Splendor, but is a unique modifier for any other buff ` - https://gbf.wiki/Seox
`Spade Trick: The boost to ATK is a unique modifier that stacks additively with Perpetuity Ring and Shield of Eternal Splendor.` - https://gbf.wiki/Caim

PR #233, opened 2019-06-05
Add supplemental damage buff + Add Covenant skills + Limit Hollowsky ( 1 per grid)

h-yasha:
- [x] Supplemental damage logic based on: https://gbf.wiki/Supplemental_Damage
- [x] Added Covenant Skills (Akasha/Hollowsky skill 2)
- [x] Limit Hollowsky per combination based on #222
> _one of the solutions i have considered (done, not in this RP however) is adding a new field, "family", and have presets for like dark opus or any weapons that have some special restrictions like only able to use 1 in each party (and some placeholders for user desired input). Could help if there were no 'easy' name matching for future._

- [x] Added a new field for Arms that only appears on specific skill for extra input.
      (currently only for Victorious Covenant and Calamitous Covenant)
- [x] Changed expectedTurn being Infinity "solution",
      doing calculations with Infinity introduces an unwanted 'overhead'.

Uncompleted/extra details:
- Need to account for Bonus DMG in supplemental damage as it applies to it separately. (Add an echo count field for Djeeta and each character?)
- [x] missing some JP/ZH translations (or correction)

**Update: (missed): translate 'debuff resistance' in results table**
Update 2: Supplemental Damage now is calculated in Graph too


src/global_logic.js
```diff
@@ -680,16 +691,90 @@ module.exports.calcBasedOneSummon = function (summonind, prof, buff, totals) {
...
+        else if (totals[key]['covenant'] === 'contentious' && taRate > 0) {
+            supplementalDamageArray['contentious'] = {};
+            supplementalDamageArray['contentious']['damage'] = Math.ceil(taRate * 100000);
+            supplementalDamageArray['contentious']['damageWithoutCritical'] = Math.ceil(taRate * 100000);
+        }
+        else if (totals[key]['covenant'] === 'deleterious') {
+            var hasCrit = false;
+            var critRate = 0;
+            for (var val in criticalArray) {
+              hasCrit = true;
```
kei-gbf: minor coding issue:
can you fix the indent spaces width 4

src/translate.js
```diff
@@ -1240,18 +1240,18 @@ var multiLangData = {
...
-        "ja": "Supplemental Damage Source",
+        "ja": "与ダメージ上昇効果のソース",
         "zh": "Supplemental Damage Source",
     },
     "Total":{
         "en": "Total",
-        "ja": "Total",
-        "zh": "Total",
+        "ja": "全額",
+        "zh": "全額",
```
kei-gbf: "全額" sound like used in money accounting.

- I don't know where it's used yet. I suggest here "ja": "合計",
- I don't know about "zh"

h-yasha: Used in results table (in show skill amount, supplemental damage table)

kei-gbf:
Add fields to save data.

- [ ] DjeetaBuffsCount
- [ ] debuffsCount

How about "buffsCount" and "enemyDebuffsCount" ?

- debuffsCount ... when new skill which effect by characters debuff, its not sure chara's or enemy's
- buffsCount ... better if each characters has own field.

kei-gbf:
> one of the solutions i have considered (done, not in this RP however) is adding a new field, "family", and have presets for like dark opus or any weapons that have some special restrictions like only able to use 1 in each party (and some placeholders for user desired input). Could help if there were no 'easy' name matching for future.

General logic sounds nice :+1:
an unique weapons is now common case.

kei-gbf:
- [x] Limit Hollowsky per combination based on #222
   - [x] Test the exclusion check with lv100

hollowsky

src/global_logic.js
```diff
@@ -680,16 +691,90 @@ module.exports.calcBasedOneSummon = function (summonind, prof, buff, totals) {
...
+            expectedCycleDamage += ougiDamage + chainBurst / buff["chainNumber"];
+
+            if (supplementalDamageArray['contentious'] != undefined) {
+                expectedCycleDamage += supplementalDamageArray.contentious.damage * expectedTurn;
+                damage += supplementalDamageArray.contentious.damage;
+                damageWithoutCritical += supplementalDamageArray.contentious.damageWithoutCritical;
+            }
 
+            var expectedCycleDamagePerTurn = expectedCycleDamage / (expectedTurn + 1.0);
+        }
```
kei-gbf:
```diff
- var expectedCycleDamage = expectedTurn * expectedAttack * damage;
+ let expectedCycleDamage = expectedTurn * expectedAttack * damage;
```

Can you change the declare keyword "var" to "let" for `expectedCycleDamage.`

Motocal used "var"  the most cases. mixing let and var without reason, may make a confuse.
However, "var" in if block make a new issue, when Infinity case expectedCycleDamage is undefined.

Change the "var" -> "let" can resolve this by expectedCycleDamage is undefined after the if-block.
we should not use the unbalanced declared variable after the scope gone.

`expectedCycleDamagePerTurn` must use "var" if it's declared in the if-block.  this is ok.

----
Another solution is make a function to calc the expectedCycleDamagePerTurn.
As I remember, there was two places which calced expectedTurn ... in `generateHaisuiData()` function.

kei-gbf:
@h-yasha 
Can you add checkbox for list items?
This PR has several topics, I want to check the progress status of review.

```md
- [ ] Supplemental damage logic based on: https://gbf.wiki/Supplemental_Damage
- [ ] Added Covenant Skills (Akasha/Hollowsky skill 2)
- [x] Limit Hollowsky per combination based on #222
- [ ] Added a new field for Arms that only appears on specific skill for extra input.
      (currently only for Victorious Covenant and Calamitous Covenant)
- [ ] Changed expectedTurn being Infinity "solution",
      doing calculations with Infinity introduces an unwanted 'overhead'.
```

src/global_logic.js
h-yasha: currently, `expectedCycleDamagePerTurn` is being calculated once in 'normal' combination, `calcBasedOneSummon`, and another time (well... times) in `generateHaisuiData`. i don't think this really warrant creating a function for them. Although, this could reduce code clutter by a 'little'.

h-yasha:
> How about "buffsCount" and "enemyDebuffsCount" ?
> 
> debuffsCount ... when new skill which effect by characters debuff, its not sure chara's or enemy's
> buffsCount ... better if each characters has own field.

this is they currently just 'thrown' in Djeeta. buff~~s~~Count in this case makes total sense.
but for debuff~~s~~Count (enemyDebuffCount), I'll move it to _buff_, as it'd make more 'sense' than being on Djeeta _and easier to access?_ in the future for special cases. However, I'll leave debuffCount on Djeeta, add enemyBuffCount to _buff_ and add debuffCount/buffCount to chara and comment them out (as they have utterly no use currently).

> Add fields to save data.
> 
> -  DjeetaBuffsCount
> -  debuffsCount

These fields are parsed from through Arm (skill_x_Detail) currently. and they do load fine (did not check compatibility)

src/global_logic.js
```diff
@@ -2762,7 +2762,7 @@ module.exports.generateHaisuiData = function (res, arml, summon, prof, chara, st
                                     haisuiBuff[l]["normalKonshin"] += 0.01 * module.exports.calcHaisuiValue("omegaKonshin", amount, slv, remainHP)
                                 }
                                 omegaKonshinIncluded = true;
-                            }
+                            }
```
kei-gbf: unnecessary space? and comment below `}//`

src/template.js
```diff
@@ -381,6 +381,12 @@ var RegisteredArm = CreateClass({
                     arm["skill2"] = "non";
                     arm["skill3"] = "non";
                 }
+            } else if (additionalKeys === "skill1Detail") {
+                arm["skill1Detail"] = this.state.skill1Detail;
+            } else if (additionalKeys === "skill2Detail") {
+                arm["skill2Detail"] = this.state.skill2Detail;
+            } else if (additionalKeys === "skill2Detail") {
+                arm["skill3Detail"] = this.state.skill3Detail;
```
kei-gbf:
```js
} else if (additionalKeys === "skill2Detail") {
```
2 -> 3 for skill3Detail

src/global_logic.js
h-yasha: cleaned.

kei-gbf:
@h-yasha 
thx. can you fix not the list has now duplicated items.
I wanted you to add only `[ ]` (checkbox)

h-yasha:
> @h-yasha
> thx. can you fix not the list has now duplicated items.
> I wanted you to add only `[ ]` (checkbox)
@kei-gbf 
like this?

src/global_logic.js
```diff
@@ -786,7 +786,7 @@ module.exports.calcBasedOneSummon = function (summonind, prof, buff, totals) {
             }
         } else {
             // Normal attack * n times
-            var expectedCycleDamage = expectedTurn * expectedAttack * damage;
+            let expectedCycleDamage = expectedTurn * expectedAttack * damage;
```

If it was in a small function, then does not care var/let. it's a narrow scope.
Now, here it's in long logic function, limit scope by the let keyword make the code strict.

one more change request.
- [ ] declare var expectedCycleDamagePerTurn outside of if-block.

```jsx
var expectedCycleDamagePerTurn;
if (expectedTurn === Infinity) {
    expectedCycleDamagePerTurn = ...
    : 
} else {
    let expectedCycleDamage = ...
    : 
   expectedCycleDamagePerTurn = ...
}
```

I found motocal follow the rule. (but not all, as you know most part uses var)

```jsx
var hp;
if (displayRealHP) {
    // Actual HP
    hp = parseInt(0.01 * (k + 1) * onedata["Djeeta"].totalHP);
} else {
    // Residual HP ratio
    hp = k + 1
}
```

----
https://eslint.org/docs/rules/no-inner-declarations

kei-gbf:
@h-yasha 
> like this?

no, there are two "Supplemental damage logic based on ..." in the list.
and other duplicated items too.

```diff
list has duplicated items
- - Supplemental damage logic based on ...
- - [ ] Supplemental damage logic based on ...
can be single, I just added `[ ]` to your list
+ - [ ] Supplemental damage logic based on ...
```

kei-gbf:
@h-yasha 
Yes, now it's fixed. thx!

h-yasha:
_for some reason it took me "this" much to actually notice the duplication_

kei-gbf:
from https://gbf.wiki/Supplemental_Damage
I will check when I confirmed the code.

- Does apply
  - [x] Normal Attack Damage
  - [ ] Normal Attack Bonus Damage ... is additional damage in Motocal?
  - [x] C.A. Damage
  - [ ] C.A. Damage Bonus Damage
  - [x] Skill Damage ... Motocal does not support
  - [x] Summon Call Damage ... Motocal does not support
  - [ ] Chain Burst Damage
- Does not apply
  - [x] Counter Damage ... Motocal does not support
  - [x] Standard Plain Damage ... Motocal does not support
  - [x] Fixed Plain Damage ... Motocal does not support
  - [x] C.A. Fixed Damage (e.g. Yodarha (SSR)) ... Motocal does not support

----
I found the formula info for Hollowsky skill2, max 30000 or 1% of enemy MHP (max HP?)
https://twitter.com/zekasyuz/status/1080843347420278784
can add to Uncompleted list. I know it's a low priority task.

h-yasha:
>  Normal Attack Bonus Damage ... is additional damage in Motocal?

yes. would have to Add a new Field? however or this. Characters can have many bonus DMG effects as buffs or passive.

>C.A. Damage Bonus Damage

Is not being Calculated _yet_ in Motocal.

> Chain Burst Damage

I missed that it does boost it. I'll update accordingly.

kei-gbf:
It does not mean to request implement full of Supplemental Damage.
it's just check list for review.

h-yasha:
- [ ] DMG Boosted is boosted by any multiplier that increases Final Damage (e.g. Seraphic multiplier).

missed. I'll add this too.

kei-gbf:
## supplemental buff + Hollowsky skill2

| key | cond | damage | damage w/o Critical | C.A | C.B |
|----|----|----|----|----|----|
| Buff | > 0 | any | any | any | any  |
| impervious | remainHP 80% | 30000 | 30000 | 30000 | 30000 |
| victorious | buffCount | x3000 | x3000 | x3000 | x3000 |
| contentious | TA | taRate x100000 | taRate x100000 | none | none |
| deleterious | critical | critRate x3000 | none | critRate x3000 | none |
| calamitous | enemyDebuffCount | x3000 | x3000 | x3000 | x3000 |

Note: each fields has damageUp affected.


| key | weapon en | weapon ja  |
|----|----|----|
| impervious | Hollowsky Spear | 虚空の楔槍 |
| victorious | Hollowsky Staff | 虚空の拝腕 |
| contentious | Hollowsky Axe | 虚空の晶塊 |
| deleterious | Hollowsky Bow | 虚空の歪弦 |
| calamitous | Hollowsky Blade | 虚空の裂剣 |

----

## Limit check

- [x] Buff ... currently positive number, 
- [x] impervious ... by remainHP, 30000 is fixed value
- [x] victorious ... debufCount max is 10
- [ ] contentious ... taRate no check in supplemental logic
- [ ] deleterious ... critRate no check in supplemental logic
- [x] calamitous ... enemyDebuffCount max is 10

I did not search the possibility of rate over 1.0
it's safe to ensure check the limit explicitly in the logic.

-----

```js
critRate = 1.0 - criticalArray[1];
```
I do not know the logic here. limit and undefined check only in review.

- I do not understand calcCriticalArray(), first I though fixed index number 1?
  if array length was short...
- var criticalRatioArray = {};
  so criticalArray[attackRatio] it's attack ratio 1 ? not index number.
- if it was attack ratio key, criticalArray[1.0] can show it's not integer index number.
- [x] undefined check by !isNaN

----
- [x] 1% enemy max HP limit is not implemented now. 

> Currently it's only being assumed that it is always maximum effect.

h-yasha:
> ```js
> critRate = 1.0 - criticalArray[1];
> ```
> I do not know the logic here. limit and undefined check only in review.
> 
> I do not understand calcCriticalArray(), first I though fixed index number 1?
> if array length was short...
> var criticalRatioArray = {};
> so criticalArray[attackRatio] it's attack ratio 1 ? not index number.
> if it was attack ratio key, criticalArray[1.0] can show it's not integer index number.

criticalArray has damage ratio as keys and ratio as values. so `1` refers to 'not crit`
image
image

kei-gbf:
- [x] Added a new field for Arms that only appears on specific skill for extra input.
(currently only for Victorious Covenant and Calamitous Covenant)

debuff-count
buff-count

- [x] a bit confuse in UI, number of weapon and number of buff/debuff fields
  - it's not issue in this PR. can be another UI improvement topic. (not change request, just note)


debuff-count2
buff-count2

- [x] when skill was changed, hide the optional field...ok


----
Those values are stored in arm `skill2Detail' field.

```json
    "armlist": [
      {
        "name": "[4凸]虚空の拝腕",
        "attack": 2660,
        "hp": 360,
        "armType": "wand",
        "skill1": "akasha-wand",
        "skill2": "victorious-covenant",
        "skill3": "non",
        "slv": 15,
        "considerNumberMin": 0,
        "considerNumberMax": 1,
        "element": "water",
        "element2": "water",
        "element3": "water",
        "skill1Detail": 0,
        "skill2Detail": "3",
        "skill3Detail": 0
      },
      {
        "name": "[4凸]虚空の裂剣",
        "attack": 3140,
        "hp": 260,
        "armType": "sword",
        "skill1": "akasha-sword",
        "skill2": "calamitous-covenant",
        "skill3": "non",
        "slv": 15,
        "considerNumberMin": 0,
        "considerNumberMax": 1,
        "element": "dark",
        "element2": "dark",
        "element3": "dark",
        "skill1Detail": 0,
        "skill2Detail": "3",
        "skill3Detail": 0
      },
```

@h-yasha 
Can you add `parseInt` to treatAddArmFromTemplate() in armlist.js

src/armlist.js
```diff
@@ -289,6 +289,10 @@ var Arm = CreateClass({
             element: 'fire',
             element2: 'fire',
             element3: 'fire',
+            skill1Detail: 0,
+            skill2Detail: 0,
+            skill3Detail: 0,
+
```
kei-gbf: no need this white spaces line

src/global_logic.js
```diff
@@ -751,8 +751,8 @@ module.exports.calcBasedOneSummon = function (summonind, prof, buff, totals) {
                 hasCrit = true;
                 break;
             }
-            if (hasCrit && !isNaN(criticalArray[1])) {
-              critRate = 1.0 - criticalArray[1];
+            if (hasCrit && !isNaN(criticalArray[1.0])) {
+              critRate = 1.0 - Math.max(0, Math.min(1, criticalArray[1.0]));
```
kei-gbf: :+1: and indent=4 pls

kei-gbf:
```js
var taRate = Math.min(1.0, Math.floor(totalTA * 100) / 100);
```
> taRate is already checked.

Ok, It's enough acceptable as PR.

----
FYI (not change request)

"Check *explicitly* in supplemental logic" is another point.
This is a review/maintenance cost.

- If modify a part of a long function
  - need to care the outer scope context, like where does the value come from etc..
    - it happen in debug time too
    - if the check was in or close to the logic. no need to search other places, often.
- If it's implemented as function
  - unit test can check the input/output limit and valid ranges.

when the code was modified, moved (by others), the logic may have different context in future.
if code was strict, it's still safe even the outer context was changed.

h-yasha:
>  a bit confuse in UI, number of weapon and number of buff/debuff fields
it's not issue in this PR. can be another UI improvement topic. (not change request, just note)

I thought about adding "↑" .
image
- Added new Field for additional selector based notes. _old field left unchanged._

kei-gbf:
> I thought about adding "↑" .

Yes. it's good if the label was on the top.

h-yasha:
- [ ] Add Hallesesena's Support Ability 
https://gbf.wiki/Hallessena

kei-gbf:
@h-yasha 
review status around expectedTurn and the supplemental
- [x] general coding check almost ok (limit and range check etc)
- [ ] logic tests

----
- [x] :+1: functional style array methods, lgtm except
  - [ ] Array.some with side effect code (change outer scope var by +=) is not good
        `some` method does not promise enumerate all elements.
        I did not see details yet, it not work as you expected,
       or even it's ok, should avoid that tricky.
- [ ] key("不壊の誓約" etc) in global_logic.js  can be English or ascii only letter?
      I want editor's keyword complete work, instead of copy/paste

h-yasha:
Array.some with side effect code (change outer scope var by +=) is not good
some method does not promise enumerate all elements.
I did not see details yet, it not work as you expected,
or even it's ok, should avoid that tricky.

~~I'll check it out.~~  I did change all of `.some` already. using `.forEach` instead. actually, there should be no `.some` (at least that I added).

> key("不壊の誓約" etc) in global_logic.js can be English or ascii only letter?
I want editor's keyword complete work, instead of copy/paste

i changed to this to not have duplicates in translate.js (well, same values just different keys)

kei-gbf:
- [x] translation key "不壊の誓約" is ok as it just a label
- [ ] supplementalDamageArray["不壊の誓約"] in global_logic.js
  I though it's internal only. I understood now it's used for display.

solution for duplicate values.
```js
// alias
multiLangData["impervious"] = multiLangData["不壊の誓約"];
```
it's ok now I see no chances to access specific key directly.

----
- supplementalThirdHitNames.forEach(function(key)
  - supplementalDamageArray[key].damage
- for (var supplementalDamageKey in supplementalDamageArray) {
  - supplementalDamageArray[supplementalDamageKey].damage

I feel it's time to separate function.. although refer a lot outer scope variables.

"key" is changed by the closest context.
for (let key in ...) does not change outer "key" but then can't write totals[key] inside.

----
- some of cases using for and if ... forEach(key, val) => ...) is useful.
- there are two loop approach about supplementalThirdHitNames include/exclude

I have ideas about supplementalDamageArray management 

- supplementalDamageArray record to have isThirdHit flag
  - .filter() can pick third hit supplemental
  - this approach can treat if other condition added, can add the flag
  - then now we can drop supplementalThirdHitNames
- var supplementalThirdHitNames = [];
  - push "修羅の誓約" if the skill exists
  - supplementalThirdHitNames.filter(key => supplementalDamageArray[key]).forEach ...
  - but this approach can't do exclude.

supplementalThirdHitNames is for manage the third hit,
when pass the supplementalDamageArray to other function,
it lose the info or have to pass supplementalThirdHitNames too.
this make code hard to maintenance, I suggest the first thirdHit flag approach.

src/global_logic.js
```diff
@@ -725,33 +727,40 @@ module.exports.calcBasedOneSummon = function (summonind, prof, buff, totals) {
...
-            var djeetaBuffCount = Math.min(10, totals['Djeeta']['buffCount']);
-            supplementalDamageArray['victorious'] = {
-                damage:  Math.ceil(djeetaBuffCount * 3000 * (1 + damageUP)),
-                damageWithoutCritical: Math.ceil(djeetaBuffCount * 3000 * (1 + damageUP)),
-                ougiDamage: Math.ceil(djeetaBuffCount * 3000 * (1 + damageUP)),
-                chainBurst: Math.ceil(djeetaBuffCount * 3000 * (1 + damageUP)),
+            let djeetaBuffCount = Math.min(10, totals['Djeeta']['buffCount']);
+            let value = Math.ceil(djeetaBuffCount * 3000 * (1 + damageUP));
+            supplementalDamageArray["凱歌の誓約"] = {
+                damage:  value,
```
kei-gbf: two spaces

src/global_logic.js
```diff
@@ -797,9 +808,9 @@ module.exports.calcBasedOneSummon = function (summonind, prof, buff, totals) {
         if (expectedTurn === Infinity) {
             expectedCycleDamagePerTurn = expectedAttack * damage;
 
-            ['contentious', 'supplemental_thirdHit'].some(key => function(key) {
-                if (supplementalDamageArray.includes(key)) {
-                    expectedCycleDamagePerTurn += supplementalDamageArray[key].damage * expectedTurn;
+            supplementalThirdHitNames.forEach(function(key) {
+            if (supplementalDamageArray[key] != undefined) {
```
kei-gbf: indent+4  and  forEach((key) => { ... }) style function for Array methods

kei-gbf:
code clean up idea (just on editor view, not tested)

- [ ] `criticalArray.length > 0` is ok?
  - if it was not so simple check, make a function hasCritical()
- [x] `supplementalDamageArray` is now same indent level with others.
- [ ] Note: `critRate` ... it's a bit complex change, I did not test this code.
      removed `hasCrit` but no need to be one-liner. here can be more simple.

```js
else if (totals[key]['covenant'] === 'deleterious' && criticalArray.length > 0) {
    let critRate = 1.0 - (isNaN(criticalArray[1.0]) ? 0 : Math.max(0, Math.min(1, criticalArray[1.0])));
    let value = Math.ceil(critRate * 30000 * (1 + damageUP));
    supplementalDamageArray["致命の誓約"] = {
        damage: value,
        damageWithoutCritical: 0,
        ougiDamage: value,
        chainBurst: 0,
    };
}
```

kei-gbf:
Object filtering sample code.
about supplementalDamageArray, supplementalThirdHitNames

https://codepen.io/anon/pen/MdNzNa?editors=0011

I set playground anyone can "Fork" and "Edit"

src/global_logic.js
h-yasha: _removed._

h-yasha:
- [ ] Supplemental Damage Buff is not being calculated.
- [ ] additional skill detail field not resetting after skill change.

kei-gbf:
- ["other", "hp_based"] given white list for exclusion (not "third_hit")
  - when new type is added, edit two places
  - [x] (not change request) it's ok acceptable, currently only two places.
  - in pickBy, type => type != "third_hit"
    - or flag approach can treat more composed situations.
       (this is unnecessary flexibility) third_hit + hp_based

:+1: implement the logic as function 
previous supplemental damage was unsafe by changes.
if others moved the code order for refactoring, calculation result might changes.
it's now strict, the code in independent function scope, not target of long function cleanup. 

----

- [x] buff count ... in main logic
- [x] enemy debuff count ... in main logic
- [x] critical ... in main logic
- [x] taRate ... in main logic
- [x] hp80% in supplemental logic
  - [ ] why only impervious checks in supplemental logic?
    - if it was in main logic, calcSupplementalDamage  can remove additionalVal

main logic ... here mean, in calcBasedOneSummon()

----

`[newDamage, undefined, newOugiDamage, chainBurstSupplemental] =`

- [ ] not good to override the `undefined`, even it's same value.
coding checker may warn this.
  - it's an acceptable cost just pass unused damageWithCritical and +0
  - I think calcSupplementalDamage can be more simple without "switch"
    - damage amount is already calculated in main logic
      the function does only object filtering and add values.

----
about the diff (not change request)

diff can show deep-green on the changed part in a line. (here on markdown highlight does not)
the changes contain unnecessary info by move+one-line.

calcChainBurst ... is not changed but it's reported as diff the code was moved.

```diff
- var chainBurst = module.exports.calcChainBurst(buff["chainNumber"] * ougiDamage, buff["chainNumber"], module.exports.getTypeBonus(totals[key].element, prof.enemyElement), chainDamageUP, chainDamageLimit);
+ var chainBurst = chainBurstSupplemental + module.exports.calcChainBurst(buff["chainNumber"] * ougiDamage, buff["chainNumber"], module.exports.getTypeBonus(totals[key].element, prof.enemyElement), chainDamageUP, chainDamageLimit);
```

I expect this diff here, but ok I understood now calc chainBurstSupplemental first.
(code is moved then reported as diff)

```diff
var chainBurst = module.exports.calcChainBurst(buff["chainNumber"] * ougiDamage, buff["chainNumber"], module.exports.getTypeBonus(totals[key].element, prof.enemyElement), chainDamageUP, chainDamageLimit);
+ chainBurst += chainBurstSupplemental;
```

h-yasha:
> ["other", "hp_based"] given white list for exclusion (not "third_hit")

Although, I'm a bit not happy with how types are passed to the function. Yes, it's just two places now. could look to define them as constants. 

> * [x]  hp80% in supplemental logic
>   
>   * [ ]  why only impervious checks in supplemental logic?
>     
>     * if it was in main logic, calcSupplementalDamage  can remove additionalVal

then I'll have to re calc it in `generateHaisuiData`, where it could simply just pass remainHP, _`k`_.

> ```js
> [newDamage, undefined, newOugiDamage, chainBurstSupplemental] =
> ```
> 
>    not good to override the undefined, even it's same value.
> coding checker may warn this.
> 
> it's an acceptable cost just pass unused damageWithCritical and += 0
> I think calcSupplementalDamage can be more simple without "switch"
> 
> damage amount is already calculated in main logic
> the function does only object filtering and add values.

Alright. I used switch, just in case in future more "types" are added.
- since it should only match once, wouldn't `break;` be better in this case? (i did this change now.)

kei-gbf:
> then I'll have to re calc it in generateHaisuiData, where it could simply just pass remainHP, k.

I see, I had not look generateHaisuiData.
it's ok ty for the explain.

> since it should only match once, wouldn't break; be better in this case? (i did this change now.)

it works ok for this case, two points 'break' pros
- if for-switch-continue was uncommon for other developers, 
  need to search the context where the `continue` is for.
  - 'break' ... is for the closest context 'switch'
  - 'continue' ... is for the outer context 'for' ... it had to read when the switch grow and long.
- need break when do something later each iteration. debug logging etc.

@h-yasha 
review is nearly finish, do you have something tasks left?

----
## Checklist
- [ ] test old save data loading, backward compatibility
  - [ ] profile: buttCount and enemyDebuffCount
  - [ ] skill1..3 Detail for weapons
  - [ ] no another fields added to save data?
- [ ] test supplement damage
  - [ ] impervious ... hp_based
    - [ ] test hp80%+ or less
    - [ ] Haisui graph
  - [ ] victorious  buff count
  - [ ] contentious taRate 
    - [ ] case check when ougi gage up -100%
  - [ ] deleterious critical
  - [ ] calamitous enemy debuff count
  - [ ] Hallesesena's Support Ability

h-yasha:
@kei-gbf 
- [ ] contentious covenant/`third_hit` type, `damageWithoutCritical` is referred to as `Single Attack Damage` (_does not reflect critical as name suggests, and **multi attack**_). Which in this case would be false presenting, as the supplemental damage is only applied to third hit. (just actually noticed)

other than this, nothing.

kei-gbf:
> Although, I'm a bit not happy with how types are passed to the function. Yes, it's just two places now. could look to define them as constants.

is it now only two cases? ["third_hit"] and ["other", "hp_based"]
Change the argument orders, pass types 1st then bind the parameter:

```jsx
// assume: function calcOtherSupplementalDamage(types, supplementalDamageArray, vals, additionalVal)
const calcThirdHitSupplementalDamage = calcOtherSupplementalDamage.bind(null, ["third_hit"]);
const calcSingleHitSupplementalDamage = calcOtherSupplementalDamage.bind(null, ["other", "hp_based"]);
```

.bind() works
- Motocal use .bind() for binding `this` context and key for React component event handlers.
- bind parameter for haisui/konshin in global_logic.test.js

bind2nd
- C++ has std::bind2nd (obsoleted in latest C++ version)
- it's easy to implement Function.prototype.bind2nd just re-arrange arguments order and call function
  - but the function call is overhead in JavaScript.
    - other languages can do this trick at compile-time
    - JavaScript .bind2nd polyfill implementation does in run-time, it's just wrapper function

----
just for comparison
move it to the last then default parameter 
supplementalDamageArray, types
```js
function calcOtherSupplementalDamage(supplementalDamageArray, vals, additionalVal, types=["other", "hp_based"])
```
can omit ["other", "hp_based"] but not ["third_hit"]

kei-gbf:
I was going to suggest .bind() approach.

Unit tests and more ideas (coding topic)
https://codepen.io/anon/pen/mZdbPN?editors=0011

NOTE: I did not check ES6 supports, at least it worked in codepen.io
but I am not sure that with Motocal's babel configuration.

h-yasha:
> I was going to suggest .bind() approach.
> 
> Unit tests and more ideas (coding topic)
> https://codepen.io/anon/pen/mZdbPN?editors=0011
> 
> * [ ]  separate to supplemental.js module
>   * [ ]  add unit tests supplemental.test.js
> * [ ]  case "hp_based": fall-through to case "other"
> * [ ]  improve additionalVal (ES6)
> 
> NOTE: I did not check ES6 supports, at least it worked in codepen.io
> but I am not sure that with Motocal's babel configuration.
> 
> Noticed in unit tests
> * switch/default case is rare to reach.
>   * that does not report single typo in supplementalDamageArray

- [x] using .bind() method.
- [x] move logic to `supplemental.js` module.
 * ? have covenant weapon skill values be in `global_const.js` rather than hard-coded in `global_logic.js`
- [x] other stuff (fall-through, improve additionalVal (ES6)

<hr>
unrelated, I installed up eslint-thingies. and noticed that there's rather a lot of variables that are declared yet unused.

image

src/supplemental.js
```diff
@@ -0,0 +1,38 @@
...
+                }
+                if (remainHP < supplemental.threshold)
+                    continue;
+                /* FALLTHROUGH */
+            case "other":
+                vals[0] += supplemental.damage;
+                vals[1] += supplemental.damageWithoutCritical;
+                vals[2] += supplemental.ougiDamage;
+                vals[3] += supplemental.chainBurst;
+                continue;
```
kei-gbf: in this case, "break" is better for switch/case. because if statement has continue.
and logging each iteration for debug. (e.g. add code to display vals changes each iteration, w/o step debugger)

src/supplemental.test.js
```diff
@@ -0,0 +1,91 @@
...
+ 
+            // default expectedTurn: 1
+            let vals = calcThirdHitSupplementalDamage(supplementalDamageArray, INITIAL_VALS(2));
+            expect(vals).toEqual([10, 10]);
+
+            // safe to pass Infinity
+            vals = calcThirdHitSupplementalDamage(supplementalDamageArray, INITIAL_VALS(2), {expectedTurn:Infinity});
+            expect(vals).toEqual([10, 10]);
+
+            // bat not -Infinity (currently, not happen in actual global_logic.js)
```
kei-gbf: my typo, I updated unit test a bit in same url.

src/supplemental.test.js
```diff
@@ -0,0 +1,91 @@
+var {calcSingleHitSupplementalDamage, calcThirdHitSupplementalDamage} = require('supplemental.js');
```
kei-gbf: const ... and add empty line to next.

`calcSingleHitSupplementalDamage` names can be like `calcSingleHitDamage` ?
in global_logic.js, use with module name prefix can show the context calc what damage.
 `supplementa.calcSingleHitDamage`

this change require describe("#calcSingleHitSupplementalDamage" ...  labels change too.

kei-gbf:
> unrelated, I installed up eslint-thingies. and noticed that there's rather a lot of variables that are declared yet unused. (well they're references? shouldn't really impact it a lot?)

Yes. it is not easy resolve those all legacy code.
no need to touch them in this PR,
it should be independent topic PR.

kei-gbf:
> ? have covenant weapon skill values be in global_const.js rather than hard-coded in global_logic.js

- or in supplemental.js
  - to be standalone for unit test.
  - it's a large file, require global_const for the small thing make big dependency.
    -  if other module require supplemental's consts, global_consts can re-produce.
  - other module does not need to know the supplemental's consts if it was private use only.

and some of code which use the consts can move to supplemental module too.

----
if it was in supplement.js then merits
- other developers can find the most of supplemental logic info in the module,
   no need to search a lot in global_logic.js/global_const.js (if they did not use IDE)
  - but no need to move all code to supplemental, global_logic already has nested if/else for skills
    - following the legacy code rules is also ok. (if use a lot outer scope variables)
- implement as function in module
  - unit test can check the piece of code
  - function scope to avoid variable name conflict. like: key2

hard code demerit. just function call overhead ...
it's acceptable cost if it not called so many times,
and require performance tuning etc. no need for now.

kei-gbf:
@h-yasha 
those are just minor coding topics, not change request.
I will soon test and finish review tonight.

src/supplemental.test.js
h-yasha: > I was going to suggest .bind() approach.
> 
> Unit tests and more ideas (coding topic)
> https://codepen.io/anon/pen/mZdbPN?editors=0011
> 
> * [ ]  separate to supplemental.js module
>   
>   * [ ]  add unit tests supplemental.test.js
> * [ ]  case "hp_based": fall-through to case "other"
> * [ ]  improve additionalVal (ES6)
> 
> NOTE: I did not check ES6 supports, at least it worked in codepen.io
> but I am not sure that with Motocal's babel configuration.
> 
> Noticed in unit tests
> 
> * switch/default case is rare to reach.
>   
>   * that does not report single typo in supplementalDamageArray

@L170 
```js
        let vals = calcSupplementalDamage(["unknown"], supplementalDamageArray, INITIAL_VALS());
```
this would require exporting the original function

h-yasha:
@kei-gbf 
> I was going to suggest .bind() approach.
> 
> Unit tests and more ideas (coding topic)
> https://codepen.io/anon/pen/mZdbPN?editors=0011

@L170 
```js
        let vals = calcSupplementalDamage(["unknown"], supplementalDamageArray, INITIAL_VALS());
```
this would require exporting the original function

kei-gbf:
> this would require exporting the original function 

Ah, Yes. unit test requires it export.

- in global_logic.js  I assumed use them with module name prefix
- in unit test, const { ... } = require style is ok, it's not so many functions.

kei-gbf:
Hollowsky Spear

Djeeta HP 100% and Vajla HP 10%

s4

HP10% but the skill is in info, is this display issue ?

Because remainHP check is in supplemental.js,
display logic also need to check remainHP ?

- [ ] Chara ATK does not contain supplemental damage?
- [ ] expected turn damage was changed HP100% 50404  -> HP10% 11872

kei-gbf:
## 更新内容 Change Log

虚空武器、与ダメ上昇系のスキル・サポアビ実装
implement Hollowsky weapons skill2, supplemental damage skill/support

もうすぐレビュー完了です。

- [x] 与ダメ上昇実装状況: status of supplemental damage
  - [x] 通常攻撃、奥義、チェインバースト ... damage, ougiDamage, chainBurst, (damageWithoutCritical)
  - [x] 追撃 ... 未対応 ... bonus(additional) damage ... currently not support
  - [x] 他アビリティダメージや奥義追撃等 ... 元カレ側が未対応 Motocal does not support others
- [x] 虚空武器第二スキル実装: status of hollowsky skill2
  - [x] 編成に一本しか入らないように ... exclusion check (1 per grid)
  - [x] 武器データ拡張: バフ、デバフ数の為 extend weapon data for buff/debuff count
    - 制限: バフの数は全員共通
  - [x] other:  虚空剣、虚空杖、虚空弓 ... 効果量はそれぞれの武器スキル依存 buff/debuff count, critical
  - [ ] hp_based: 虚空槍 HP8割チェック ... remainHP 
    - [ ] 対応待ち: HP80%未満の場合のスキル表示
  - [x] third_hit: 虚空斧 ... 連撃のみ有効。奥義には載らない ... taRate
- [ ] ハレゼナ サポアビ実装 third_hit


NOTE: チェックは確認済項目。Checkbox is for my confirmation, not mean support/implement.

h-yasha:
- Added `.filter` to supplementalDamageArray in `result.js` 
  - ~~? create a function in `supplemental.js` to return the "sorted", and "filtered" array.~~
  - [x] add a boolean to each supplemental damage item in the array, as in:
```diff
case "other":
    vals[0] += supplemental.damage;
    vals[1] += supplemental.damageWithoutCritical;
    vals[2] += supplemental.ougiDamage;
    vals[3] += supplemental.chainBurst;
+   supplemental.show = true;
    break;
```

h-yasha:
> Chara ATK does not contain supplemental damage?
image

unless i'm mistaken, Chara ATK (`totalAttack`) should be similar to ingame estimated damage. which does not contain it.

kei-gbf:
how about `show: remainHP >= threshold` in global_logic.js
or the first approach was better.

- there are some timings the calc supplemental damage function is called
  - if haisui graph is called ... when the skill info is update (auto update on/off)
  - the show flag is changed by the timing, that make hard to test all situations
- side effect for damageArray (`supplemental.show = true`) make test harder too.

----

result.js

I did not check the code.

side effect (total counting) in map() is a kind of abuse of map. although legacy code has that a lot.
in theory, same input should always return same output, (important for unit test)
but the side effect break this rule by changes the state.
this mean, program result change by the calling order, problem when run tests concurrency.

This issue is not a problem currently. works ok at least.
code quality topic. it's just report for information.

Demo code (lodash and JavaScript)
https://codepen.io/anon/pen/GbRbOr?editors=0011

- transform object to [key,value] array
- filtering hp_based and low remainHP case
- sorting
- sum total damage

----

`const info= supplemental.collectSkillInfo(supplementalDamageArray, {remainHP: 0});`
- info.keys ... array of key
- info.values ... array of damage
- info.total ... totalSupplementalDamage

// NOTE: change remainHP value to actual character's.
// NOTE: not tested.

```js
function collectSkillInfo(damageArray, {remainHP:1.0}={}) {
    const isAvailable = ([key, val]) => !((val.type == "hp_based") && (remainHP < val.threshold));
    const xs = Object.entries(damageArray).filter(isAvailable).sort();
    return {
        keys: xs.map(([key, val]) => key),
        values: xs.map(([key, val]) => val.damage),
        total: xs.reduce((total, [key,val]) => total + v.damage, 0),
    };
}
```

for .show flag
`const isAvailable = ([key, val]) => val.show;`

h-yasha:
that looks very... compact. it works (tested. not thoroughly). and also solves display inconsistency,
> - Supplemental Damage table shows with no actual data when HP Based (Impervious Covenant) is disabled.
> - While, Victorious Covenant and Calamitous Covenant show up and display 0.
> - lastly, Contentious Covenant (and Buff), does not show the table when taRate is 0. (desired outcome for all?)

~~It'll take me a while to comprehend that~~

kei-gbf:
@h-yasha 
can you remove "show" flags.

- set show flag in global_logic.js when add to supplementalDamageArray
  - [ ] if auto update off (not tested) this may require extra tests
- set show flag in calc damage ... timing issue, haisui graph may effect to this.
- check remainHP on demand is better

----

> show damage 0

sorry, no needed. now any jobs for Djeeta available the skill and show damage 0
even weapon type was different.

kei-gbf:
hide the skill info when damage is 0, maybe easy solution

kei-gbf:
- [x] test Hallesesena's with Hollowsky Axe
  - [x] test TA buff to taRate 0..100 % case

----
missing fav1,fav2 check for supplementalDamageArray add?

h-yasha:
> missing fav1,fav2 check for supplementalDamageArray add?

fav arm is explicitly relevant for covenant weapon skill. matched in `addSkilldataToTotals`.

<hr> 

_draft:_
- [x] Hide supplemental damage when value is 0.
- [ ] details on what type of supplemental damage it is ("normal", on crit, or third hit) >>> currently only for hallessena supabi, contentious covenant, and deleterious covenant (first two are third hit, and last is crit only)

kei-gbf:
> lastly, Contentious Covenant (and Buff), does not show the table when taRate is 0. (desired outcome for all?)



- [ ] Contentious hide, but Hallesesena's support stay.
  - hide damage 0 info ... add val.damage check to isAvailable
  - or hide table if (supplementalInfo.total != 0) ...

h-yasha:
> > lastly, Contentious Covenant (and Buff), does not show the table when taRate is 0. (desired outcome for all?)
> 
> 
> 
> * [ ]  Contentious hide, but Hallesesena's support stay.
>   
>   * hide damage 0 info ... add val.damage check to isAvailable
>   * or hide table if (supplementalInfo.total != 0) ...

Done. just doing second task

h-yasha:
@kei-gbf need to check on translations.

src/translate.js
```diff
@@ -1263,6 +1263,26 @@ var multiLangData = {
...
+        "en": " (Applies to third hit)",
+        "ja": " (3回目の攻撃に)",
+        "zh": " (Applies to third hit)",
+    },
+    "supplemental_on_critical": {
+        "en": " (Applies to critical hit)",
+        "ja": " (クリティカル攻撃に)",
+        "zh": " (Applies to critical hit)",
+    },
+    "supplemental_hp_based": {
```
kei-gbf: if need threshold in display text,
make a skill->skill name function
and place-holder for the threshold info.

```json
"en": "Applies when remainHP is {threshold}%+", 
```

```js
case "hp_based":
     return intl.translate("supplemental_hp_based", locale).replace("{threshold}", supplemental.threshold);
```

src/result.js
```diff
@@ -1305,14 +1305,14 @@ var Result = CreateClass({
...
-                        if ("keys" in supplementalInfo && "values" in supplementalInfo && "total" in supplementalInfo) {
+                        if ("keys" in supplementalInfo && "values" in supplementalInfo && "total" in supplementalInfo && supplementalInfo.total > 0) {
                             supplementalDamageInfo.push(
                                 <table key={key + "-supplementalDamageTable"} className="table table-bordered" style={{"marginBottom": "0px"}} >
                                     <thead>
                                         <tr>
                                             <th className="bg-success" style={{"fontSize": "10pt"}}>{intl.translate("与ダメージ上昇効果のソース", locale)}</th>
                                             {supplementalInfo.keys.map( function (v, ind) {
-                                                return <th key={ind} className="bg-success" style={{"fontSize": "10pt"}}>{intl.translate(v, locale)}</th>
+                                                return <th key={ind} className="bg-success" style={{"fontSize": "10pt"}}>{intl.translate(v[0], locale) + intl.translate(("supplemental_"+v[1]), locale)}</th>
```
kei-gbf: (v, ind) -> ([key, type], ind)

src/translate.js
kei-gbf:
```json
"ja": "残りHPが {threshold} 以上の時",
```

src/translate.js
```diff
@@ -1263,6 +1263,26 @@ var multiLangData = {
         "ja": "サポアビ",
         "zh": "Support Ability",
     },
+    "supplemental_other": {
```
kei-gbf: `type` lose the info here which buff count or enemy debuff count.

src/result.js
kei-gbf: Reason: 
v[0] and v[1] is hard coded magic number, Readers need to search what was the values. 
[key, type] naming info can give the hints.

src/translate.js
h-yasha: point of this is just to specify to what the damage is applied to. that's just too much details (check the wiki, or 'save' tab, _which probably need to be changed at some point_).

src/result.js
```diff
@@ -1305,14 +1305,14 @@ var Result = CreateClass({
 
                         var supplementalDamageInfo = [];
                         const supplementalInfo = supplemental.collectSkillInfo(skilldata.supplementalDamageArray, {remainHP: m.data[key].remainHP});
-                        if ("keys" in supplementalInfo && "values" in supplementalInfo && "total" in supplementalInfo) {
+                        if ("keys" in supplementalInfo && "values" in supplementalInfo && "total" in supplementalInfo && supplementalInfo.total > 0) {
```
kei-gbf: it's safe to remove those property checks, when empty object was given
supplementalInfo.keys, values is [] (empty array) map() will just ignore. total is default value 0.

kei-gbf:
- [ ] add unit test for collectSkillInfo()
  https://codepen.io/anon/pen/agzMOR?editors=0010
- [ ] remove "keys", "values", "total" property check
- [ ] fill those empty translation fields

src/result.js
kei-gbf: - [ ] <th key should be unique?
  - https://reactjs.org/docs/lists-and-keys.html#keys
- [ ] arrow function can make omit "function" "return"

h-yasha: >  \<th key should be unique?

it's unique, is it not?

>  arrow function can make omit "function" "return"
 
?

h-yasha:
- [ ] update supplemental.test.js to reflect on latest changes. woops

h-yasha:
> * [ ]  fill those empty translation fields
left empty intentionally.

src/result.js
kei-gbf: > arrow function can make omit "function" "return"

I should type "arrow-function"
it used when write a function in a expression.
```js
{supplementalInfo.keys.map((v, ind) => <th>...</th>)}
```
used in map(), .filter() usually.

> it's unique, is it not?

it's ok. seem that was safe case to use index as keys.
I thought "key" of supplementalInfo could the unique.

> Keys Must Only Be Unique Among Siblings
> Keys used within arrays should be unique among their siblings. However they don’t need to be globally unique. We can use the same keys when we produce two different arrays:

index is ok here. now our contents is static.
Document said that may break unique when list item change dynamically.

> We don’t recommend using indexes for keys if the order of items may change. This can negatively impact performance and may cause issues with component state.

https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318
> When all of them are met, you may safely use the index as a key.

- [x] 1. the list and items are static–they are not computed and do not change;
- [ ] 2. the items in the list have no ids;
  - or "key" of supplementalInfo can be the key for React <th>
    it's unique keys of supplementalDamageArray
- [x] 3. the list is never reordered or filtered.
  - it filtered hp-based, but not dynamic filtering ... it's re-calc and generate with new indexes.

kei-gbf:
```
expect(supplementalInfo.headers).toEqual([
    ["B", "hp_based", null],
    ["C", "third_hit", null],
    ["D", "other", null],
    ["E", "on_critical", 50]
]);
```
it's a maintenance cost to check full data,
when change the test data source, need to update another expected data.

in the expect(headers) we just want to test isAvelable filtering works.

```js
let toKey = v => v[0]; // toKey (["B", "hp_based", null]) => "B"
expect(supplementalInfo.headers.map(toKey)).toEqual(["B", "C", "D", "E"]);
```
NOTE: if the temporary "toKey" function something heavy task, need another tests
it's simple and short for this case.

or if don't want write code like v[0]
info provide "keys" and "headers", then use keys field for the test
headers for result.js a bit redundant but not much run-time cost, it used only once in result.js

src/supplemental.js
```diff
@@ -12,6 +14,10 @@ function calcSupplementalDamage(types, damageArray, vals, {remainHP = 1.0, expec
                 if (remainHP < supplemental.threshold)
                     continue;
                 /* FALLTHROUGH */
+            case "boss_debuff_based":
+                /* FALLTHROUGH */
+            case "djeeta_buff_based":
+                /* FALLTHROUGH */
```
kei-gbf: This FALLTHROUGH chain may not work as you expected
boss_debuff_based will run all  boss_debuff_based, on_critical and other
or require ensure double type check in the case
case "djeeta_buff_based": if (type was "djeeta_buff_based") { // to avoid anoter type

kei-gbf: I see now, it currently nothing to do in other fall through cases.

if/else is flexible for this use?

- if (hp_based and remainHP) continue // only filtering
- if (third_hit) ...
- else ... hp_based also add damage here
- no need to care unknown case since it filtered by types.includes

h-yasha: I changed their type from "other" to specific (boss_debuff_based/etc) to avoid introducing more values/variables. these are supposed to be same as "other". (and as you mentioned. they're already filtered by include, or it should not get any unexpected values.)

kei-gbf:
> left empty intentionally.
can you add that to comment.

Readers does not know the context how it used.
they may think it's missing or forget.

kei-gbf:
> it's a maintenance cost to check full data,
> when change the test data source, need to update another expected data.

self reply.
in this point, values check in "test damageArray remainHP 100%,50%" can be removed.
values are not unique, since total is checked, no particular  reason to check values.
it's ok currently, but when test data grow and became large test.

kei-gbf:
how about add "label" new field? now "type" has two role that make code complex
- current "type" code does not need to change, ["other", "hp_buff", "third_hit"]
- label is for display info

kei-gbf:
## Separate filter and damage calculation.

```js
// filtering only, type specific code should be here
if (!types.includes(supplemental.type))
    continue;
if (supplemental.type ===  "hp_based") {
    if (!("threshold" in supplemental)) {
        console.error("Missing HP threshold in: " + supplemental);
        continue;
    }
    if (remainHP < supplemental.threshold)
        continue;
}

// calc damage only, either if/else  or  switch/case "third_hit"/default
if (supplemental.type === "third_hit") {
    if (expectedTurn === Infinity) {
        expectedTurn = 1;
    }
    vals[0] += supplemental.damage;
    vals[1] += supplemental.damage * expectedTurn;
} else {
    vals[0] += supplemental.damage;
    vals[1] += supplemental.damageWithoutCritical;
    vals[2] += supplemental.ougiDamage;
    vals[3] += supplemental.chainBurst;
}
````

- [x] sample case, type was like hp_based & third_hit
  hp_based & third_hit code are still reusable.
  - it's hard in single switch, that mixed two kind of tasks filter/calc damage.

**Comment by kei-gbf** [comment]
@h-yasha 
I explain why the fall through chain is not suit for this case.

```js
switch (type) {
case A: // fall through
case B: // fall through
case C: // fall through
}
```

the relation is B include C ... mean B is extended of C
but not ok for A. A include B and C ... A should not have relation with B.
(replace with actual types: `"hp_based"` should not have relation with `"boss_debuff_based"`)
it's ok if A was super set of B and C, but not in this case.

If there are many types added in future, then fall through approach will break soon.
I think it's a easy and temporary solution.
OOP can handle this logic more flexible by extends classes. (but heavy approach for now)

kei-gbf:
```js
switch (type) {
case A: // fall through to C
case B: // fall through to C
case C: break;
}
```

Reader can notice B should not have type specific code.
but no much reason to use fall through approach,
they can be just default case.

other idea is:
```js
switch (supplemental.type) {
case "third_hit": break;
case "hp_based": // fall through to default
default: break;
}
```

- move third_hit to top, then we can use default case for other.
- comment "fall through to ..." still match with eslint's rule,

src/supplemental.js
```diff
@@ -53,8 +53,10 @@ function collectSkillInfo(damageArray, {remainHP = 1.0}={}) {
 
 function tableHeader([key, type, additionalVal], locale) {
     var str = intl.translate(key, locale);
-    str += intl.translate(("supplemental_" + type), locale);
-    str = str.replace("{value}", additionalVal == null ? "" : additionalVal);
+    if (type != undefined) {
+        str += intl.translate("supplemental_" + type, locale);
```
kei-gbf: `intl` is global environment value, it's unnecessary dependency for supplemental module.
can this change pass translated value?

or that function should be defined in result.js

kei-gbf: I did not look the place-holder replacement.

- supplemental unit test does not need to check translation, that translate's territory.
  - generally, unit test should not have outer dependency like "intl", it's environment value for tests.
    when change the environment that effect to tests.
  - unit test does only logic test. replace place-holder is common task, not supplemental specific
    the supplemental specific part is only the place-holder's key name, or the data record types.
- at this point, tableHeader function can be in result.js  that already has a lot dependencies.
  - this just aim keep supplemental module simple and less dependency.
  - the test for tableHeader would be an integration test. (using same Jest framework is ok)

kei-gbf:
## new tasks
- [ ] unique key for list
- [ ] rearrange switch cases
- [ ] improve module dependency of the place holder function for table header

I will make PR. it's hard to explain with my English.

kei-gbf:
just note, naming a bit conflict
"additionalVal" Motocal already use `additionalDamage` in global_logic.js

- additional for supplemental info
- additional for damage

if anyone had better rename idea, rename this to avoid
when search "additional" in editor, hits both.

kei-gbf:
@h-yasha 
- [ ] I think this is the last task https://github.com/MotocalDevelopers/motocal/pull/233#issuecomment-501577669
  comment on empty translation

----
Update summary https://github.com/MotocalDevelopers/motocal/pull/233#issuecomment-500819134

h-yasha:
_(I got a bit busy, nonetheless)_

> just note, naming a bit conflict
> "additionalVal" Motocal already use `additionalDamage` in global_logic.js
> 
> * additional for supplemental info
> * additional for damage
> 
> if anyone had better rename idea, rename this to avoid
> when search "additional" in editor, hits both.

~~value2~~ wouldn't `value` be ok? since it's mainly the `damage` that is used for calculation. otherwise, `extraValue`.

kei-gbf:
> wouldn't value be ok? since it's mainly the damage that is used for calculation. otherwise, extraValue.

ok, as it replaced with {value}, plain name
supplemental skill info had "values" field that collect damage, ok if no much confuse.

I worried was mislead of the supplemental/additional,
when want to follow additional damage code.

h-yasha:
I went for `extraValue`,
it would be a little bit confusing to do `value`
```js
let djeetaBuffCount = Math.min(10, totals['Djeeta']['buffCount']);
let value = Math.ceil(djeetaBuffCount * 3000 * (1 + damageUP));
supplementalDamageArray["凱歌の誓約"] = {
    damage: value,
    damageWithoutCritical: value,
    ougiDamage: value,
    chainBurst: value,
    type: "djeeta_buff_based",
    extraValue: djeetaBuffCount,
}
````

kei-gbf:
与ダメ上昇＆虚空武器第2スキル実装のレビュー完了です。
更新内容の要約は https://github.com/MotocalDevelopers/motocal/pull/233#issuecomment-500819134

制限は以下の通り
- アビダメや奥義追撃には未対応 (Motocal が対応してない)
- 通常攻撃の追撃にも未対応
- 虚空杖 バフの数は全員共通
- 上限の敵最大HP1% は未実装

h-yasha:
@kei-gbf 
> * 虚空杖 バフの数は全員共通

if this is reference to [Victorious Covenant](https://gbf.wiki/Hollowsky_Staff)
>Supplemental damage to staff- and harp-specialty allies' attacks based on MC's number of buffs (Damage cap: 30,000)

It is based on MC number of buffs for everyone.

kei-gbf:
@h-yasha 

Yes. it's just explain the current implementation
the amount refer to Djeeta's buff count

kei-gbf:
@h-yasha 

I got your point, Yes I should not say "everyone" that only stuff, harp allies
I will improve the expression.

PR #201, opened 2019-05-16
Add preset parameter

kei-gbf:
This patch implements two another ways to load JSON data.

  * preset=HASH loads base64 encoded json data
      Data body is read from location.hash (javascript)
      Server access log will not store the long data.
  * preset=name loads ./json/name.json file
      "name" is checked \w\-\_ letters only

Internal changes:

  * loadDatachar (decoded json text)
    data load function, the original code is moved here.
  * getDatacharByFragment (preset=HASH)
  * getDatacharByURL (preset=name)
    load raw json text, no base64 encode.
  * getDatacharById (id=num)

PR #240, opened 2019-06-06
Add Uplift

h-yasha:
#62 
> バフ：高揚

<hr>

- Uplift effect Buff for Individual.
- Uplift effect Buff for Party.

_Note_: while exact uplift effect does not stack. special effects that could has uplift effect like Linkage (SS Grea) does stack with normal uplift effect.


kei-gbf:
- [x] the buff effect seem works ok
- [ ] can you add display the buff info her characters
      currently, users has no ways to check the value change
  -  implement: display buff info in result.js, pass value via `coeffs` in global_logic.js
- [x] buff range 0...100 check ok
  - [ ] but total 200+%

uplift200

h-yasha:
- [x] add display element (skill amount) for uplift

<hr>

> * [x]  buff range 0...100 check ok
>   
>   * [ ]  but total 200+%

this is fine. could represent values 100%+ (like in Kengo) and it does not harm the calculation.
could limit it for 200% maximum.

OrkunKocyigit:
I think one of the issues we skip is that game always runs down ougi gage. So even 11.99 becomes just 11. Right now we are assuming it doesn't. It is not in scope of this change request but should be considered in future

h-yasha:
> I think one of the issues we skip is that game always runs down ougi gage. So even 11.99 becomes just 11. Right now we are assuming it doesn't. It is not in scope of this change request but should be considered in future

it actually runs it up.
https://gbf.wiki/Character_Extended_Mastery_Perks#chargebar

kei-gbf:
- [ ] Another point, the buff's recast and interval.

It's easy to keep the buff with other gage up or DA/TA boost buff.

however, not just uplift buff. it's at least require Djeeta DA x 2T (+ other character ougi) ?
True Purity Sunblade ougi Buff 3T include ougi turn
- 3T * 10 % Uplift
- 2T * 24% Djeeta DA with master bonus
- can expect additional 30% if other character had ougi
when DA/TA ratio was low, can't keep uplift buff always.

This can be extra topic, if it's ok please solve conflict.

----
(just share info)
- [ ] buff's interval (the real effect may not be forever usually)
- [ ] Djeeta's master bonus (+1 per single hit)
- [ ] Chain burst
  - e.g. charge bar 100/100/71/71 can be 4 Chain with boost buff (but not uplift?)
  - MEMO: I had played in trial abuse of uplift with Xeno fist ougi
     https://github.com/kei-gbf/GBFArkabOugiTALoop
  - youtube: https://www.youtube.com/watch?v=Z2-zwqoKfCI&feature=youtu.be

Motocal does not calc it so complex?
and I know it's not simple formula, need simulate to calc those.

----
another view point

(premise) IMO, people use Motocal for weapon grid comparison mainly.
for another use case, there are another tools 1TK simulator etc (with google spreadsheet)

- how much development resource we can use for this task,
  after the implement completed the charge bar simulation.
  what benefit for the weapon comparison. (it does not mean no-need, but priority)
  The demerit is increase unnecessary maintenance cost, if it was low priority.
- or grow to be calc damage "speed" (for multi MVP)

I am not the project author, I won't decide the project purpose.
just keep in mind a bit.

h-yasha:
>   * e.g. charge bar 100/100/71/71 can be 4 Chain with boost buff (but not uplift?)

_Uplift_ effect is applied at the end of the turn.

> another view point
> 
> (premise) IMO, people use Motocal for weapon grid comparison mainly.
> for another use case, there are another tools 1TK simulator etc (with google spreadsheet)
> 
> * how much development resource we can use for this task,
>   after the implement completed the charge bar simulation.
>   what benefit for the weapon comparison. (it does not mean no-need, but priority)
>   The demerit is increase unnecessary maintenance cost, if it was low priority.
> * or grow to be calc damage "speed" (for multi MVP)
> 
> I am not the project author, I won't decide the project purpose.
> just keep in mind a bit.

charge bar boosts/etc should effect grid calculation rather well, for example:
- [Xuanwu Shellfists](https://gbf.wiki/Xuanwu_Shellfists), [Blue Sphere](https://gbf.wiki/Blue_Sphere), etc...

kei-gbf:
I will finish this review, loading test done.

- [x] test remove uplift fields in JSON and load it back in new page.
  - no JavaScript error by undefined fields.

----
## Known issue for Loading old save data (won't fix)

I note this for share info. it talked in another PR comments b4.
(I can't find which number now)

- [x] Load from server is ok, no problem.
- [ ] Load JSON file ... if the new field was changed and load old file which field was undefined
     will not override current value.

     but it's not new issue. another fields recently added also has this. except:
     "enemyDefense" does ad-hook re-initialize in content.js but not smart solution.
     This approach will conflict with #201, that I am trying partial loading feature.
     (so I want the solution be reset state before loading, not re-initialize loading data.)

PR #250, opened 2019-06-10
Add Critical Buff (old PR #245) and Sun-Touched Paraside

h-yasha:
_Recreate Branch. 'cuz whoopsie._
_old PR: #245_

>  バフ：クリティカル

#62 

<hr>

- Add [Sun-Touched Paraside](https://gbf.wiki/Alanaan#Sun-Touched_Paradise) ([烈日の楽園](https://gbf-wiki.com/index.php?%A5%A2%A5%E9%A5%CA%A5%F3%20%28SSR%29#o0051303)).

- ~~[ ] Add Critical Buff to Summons. #42 #61~~ Pointless. Could Add a fixed selection of what is already available (super buncles, and others if there's)... But Party Critical Buff do enough.

Edit:
- ~~[ ]~~ Implement Sun Touched Paradise in Damage Simulator.
    - I would leave this for now. toggling it does reflect changes in damage simulator as intended. Ideally, would add it as a state as it does lasts only [8T/12T](https://gbf.wiki/Alanaan). However, current damage simulator is not suitable for changes like this.


src/global_logic.js
```diff
@@ -598,14 +598,24 @@ module.exports.calcBasedOneSummon = function (summonind, prof, buff, totals) {
...
+            var normalOtherCriticalArray = totals[key]["normalOtherCritical"].concat(LBCriticalArray, EXLBCriticalArray, normalOtherCriticalBuffArray, totals[key]["criticalBuff"], buff["criticalBuff"]);
 
             var criticalArray = module.exports.calcCriticalArray(totals[key]["normalCritical"], totals[key]["magnaCritical"], normalOtherCriticalArray, totalSummon);
-            var criticalRatio = module.exports.calcCriticalRatio(criticalArray)
+            var criticalRatio = module.exports.calcCriticalRatio(criticalArray);
         } else {
             var damageUP = 0.0;
-            var criticalArray = {};
-            var criticalRatio = 1.0
+            if (prof.retsujitsuNoRakuen) {
```
kei-gbf: Can you consider ` else if (prof.enemyElement == "non-but-critical" || prof.retsujitsuNoRakuen) {`
to reduce the clone code.

src/global_logic.js
```diff
@@ -598,14 +598,24 @@ module.exports.calcBasedOneSummon = function (summonind, prof, buff, totals) {
...
+                var LBCriticalArray = getLBCriticalArray(totals[key]["LB"]);
+                var EXLBCriticalArray = getEXLBCriticalArray(totals[key]["EXLB"]["Critical"]);
+                var normalOtherCriticalBuffArray = totals[key]["normalOtherCriticalBuff"];
+                var normalOtherCriticalArray = totals[key]["normalOtherCritical"].concat(LBCriticalArray, EXLBCriticalArray, normalOtherCriticalBuffArray, totals[key]["criticalBuff"], buff["criticalBuff"]);
+
+                var criticalArray = module.exports.calcCriticalArray(totals[key]["normalCritical"], totals[key]["magnaCritical"], normalOtherCriticalArray, totalSummon);
+                var criticalRatio = module.exports.calcCriticalRatio(criticalArray);
+            } else {
+                var criticalArray = {};
+                var criticalRatio = 1.0
```
kei-gbf: missing `;`

src/translate.js
```diff
@@ -426,6 +451,31 @@ var multiLangData = {
...
+        "zh": "Special Effects",
+    },
+    "烈日の楽園": {
+        "en": "Sun-Touched Paradise",
+        "ja": "烈日の楽園",
+        "zh": "烈日の楽園",
+    },
+    "烈日の楽園説明": {
+        "en": "All Allies gain 25% normal buff, and can deal critical hits regardless of element. (Alanaan)",
+        "ja": "対象の属性に関係なくに味方のクリティカル発生し、味方の通常攻刃わ25%UP。(アラナン)",
```
kei-gbf:
```diff
- "対象の属性に関係なくに味方のクリティカル発生し、味方の通常攻刃わ25%UP。(アラナン)"
+ "対象の属性に関係なく味方のクリティカルが発生し、味方の通常攻刃25%UP。(アラナン)"
```

src/global_logic.js
```diff
@@ -1227,6 +1238,7 @@ module.exports.getTotalBuff = function (prof) {
     totalBuff["zenithChainDamageLimit"] += zenithChainDamageLimit[prof.zenithChainDamageLimitBonus] != undefined ? zenithChainDamageLimit[prof.zenithChainDamageLimitBonus] : 0;
     totalBuff["zenithElement"] += zenithElement[prof.zenithElementBonus] != undefined ? zenithElement[prof.zenithElementBonus] : 0;
     totalBuff["zenithDamageLimit"] += zenithDamageLimit[prof.zenithDamageLimitBonus] != undefined ? zenithDamageLimit[prof.zenithDamageLimitBonus] : 0;
+    totalBuff["criticalBuff"] = prof.criticalBuff != undefined ? prof.criticalBuff : [];
```
kei-gbf: (Not change request, just coding info)
`totalBuff["criticalBuff"] = prof.criticalBuff || [];`

could be short way (if criticalBuff was non-number,non-string object)
however, other code are checking undefined, it's ok follow legacy code rule.
(if change those all legacy code, I want refactoring topic, because that make extra tests costs)

OrkunKocyigit:
This PR breaks my old save. Can you check it?
[Save.txt](https://github.com/MotocalDevelopers/motocal/files/3317707/Save.txt)

kei-gbf:
I was going to just note (no need to start this review soon) it's ok when you finished #263 
there are a lot PR now.

----
in render() should not change state?
```js
if (this.state.criticalBuff[i] == undefined) 
    this.state.criticalBuff[i] = {"value": 0.0, "attackRatio": 0.0};
```
```js
if (this.state.criticalBuff.length != this.state.criticalBuffCount)
    this.state.criticalBuff = this.state.criticalBuff.slice(0, this.state.criticalBuffCount);
```

logic in view, may make unexpected timing. e.g. if auto update was off.
I think those logic should be in event when count was changed. 

----
@OrkunKocyigit 
> This PR breaks my old save. Can you check it?

which version the save data was made?
is it because Normal Critical changed like Magna critical ? or only in this PR?

OrkunKocyigit:
I'll say around 2 weeks old. It works with current master.

kei-gbf:
@OrkunKocyigit 

thx, I will try debug this.

> is it because Normal Critical changed like Magna critical ?
was not affect to save data (save data store skill id). so only internal change.

in calcCriticalArray() function
normalOtherCritical contained undefined elements

[{...}, {...}, {...}, undefined]

each item has "value", "attackRatio" but undefined not.

- [ ] First, Skip the undefined item is easy solution
- [ ] and search where the undefined came from.

----
Save data compatibility checks

- new fields in this PR
  - [x] retsujitsuNoRakuen 
  - [x] criticalBuff
  - [x] criticalBuffCount
  - other?

Steps:
- open Motocal page (data are all default)
- "Save" tab, input data name then "Save" and output JSON
- edit JSON data (remove those added new fields)
- reload page (because known issue, loading same data name does not refresh view)
- "Import" JSON data and "Load"

was no problem in this PR.
at least it was not so simple missing initialize fields.

kei-gbf:
I tried "Earth" in the Save.txt

```js
var normalOtherCriticalArray = totals[key]["normalOtherCritical"].concat(
  LBCriticalArray,
  EXLBCriticalArray,
  normalOtherCriticalBuffArray,
  totals[key]["criticalBuff"],
  buff["criticalBuff"]);
```
(just expand arguments in comment, the original code was single long line)

```js
// console.log(normalOtherCriticalArray) displays:
[{...}, {...}, {...}, undefined]
```

- [x] totals[key]["normalOtherCritical"]
- [x] LBCriticalArray
- [x] EXLBCriticalArray
- [x] normalOtherCriticalBuffArray
- [ ] totals[key]["criticalBuff"] <- undefined
- [x] buff["criticalBuff"]

Solution would be: either
- [ ] `totals[key]["criticalBuff"]  || []` <- if it was `undefined` then pass empty array
  - there are several places which take totals[key]["criticalBuff"]
- [ ] check `undefined` item in calcCriticalArray function

src/profile.js
```diff
@@ -254,6 +267,42 @@ var Profile = CreateClass({
     },
     render: function () {
         var locale = this.props.locale;
+        //Generate Critical Buff Fields
+        var criticalBuffRender = [];
+        if (this.state.criticalBuff.length != this.state.criticalBuffCount) this.state.criticalBuff = this.state.criticalBuff.slice(0, this.state.criticalBuffCount);
+        for (var i = 0; i < this.state.criticalBuffCount; i++) {
+            if (this.state.criticalBuff[i] == undefined) this.state.criticalBuff[i] = {"value": 0.0, "attackRatio": 0.0};
```
kei-gbf: - [ ] this "if" statement does not make indent.
- [ ] change state in render() view ... should be in event count changed timing
- [ ] can be function?(or react component) which take parameter criticalBuff, personalCriticalBuff

h-yasha:
- [x] remove redundant code in global.js and a missing semicolon
- [x] identitation in profile.js
- [x] jp text for 烈日の楽園説明

- Fix backward compatibility:
    - [ ] check undefined item in calcCriticalArray function

<hr>

> * [ ]  this "if" statement does not make indent.
> * [ ]  change state in render() view ... should be in event count changed timing
> * [ ]  can be function?(or react component) which take parameter criticalBuff, personalCriticalBuff

I'm not well versed in react (if anything, i know nothing)... regardless, I tried to have the "render" be a property and be generated when changing the count (in event handler). However, this would break save/load completely, ~~_(i don't exactly recall what the error exactly was been rather a long time since then. something invalid react object...)_~~

src/profile.js
h-yasha: > [ ]  this "if" statement does not make indent.

it's a oneline-er.

kei-gbf:
empty-tooltip

- [ ] empty translation show blank tooltip
- [ ] reduce clone codes (criticalBuff to component) chara, profile has total 3 similar codes
  - [ ] I will PR this, make it React component
  - stage: sample, https://codepen.io/anon/pen/PrjvyV?editors=0011
    - [ ] display float number in "value" or "attackRatio", need round?
      try 14,28,55-58 ... that show 14.000000000...

kei-gbf:
>  check for invalid summon?

I think currently no need, it only magna and zeus
we will check it when epic series summon (critical main aura) add new field to summon.

h-yasha:
> * [ ]  empty translation show blank tooltip

ah... i missed that.

> * [ ]  need max count?

i don't really think that's needed. there's not actually fixed amount of critical sources (also could practically hold any critical effects like if some support abilities are not implemented.

kei-gbf:
## maxCount
> i don't really think that's needed. there's not actually fixed amount of critical sources (also could practically hold any critical effects like if some support abilities are not implemented.

but no need too much count, that can make resource attack by modify the count field.
(if that count was too large amount, save data size and browser resource usage by too much list items)
I set in my PR maxCount: 20, if need more or less you can change it later.

kei-gbf:
I have one more topic,
reduce clone code and make code more strict.

it's not tested code, I will explain about the variable scope and why it's important.

```js
var damageUP = 0.0;
if (totals[key]["typeBonus"] == 1.5) {
    // Supplemental damage rise support ability does not overlap with Tenshi skill (the strongest effect overwrites the lesser)
    damageUP = Math.max(totals[key]["tenshiDamageUP"], totals[key]["charaDamageUP"]);
    damageUP += 0.01 * totalSummon["tenshiDamageUP"];
    damageUP += totals[key]["charaUniqueDamageUP"];
}

var criticalArray = {};
var criticalRatio = 1.0;
if (totals[key]["typeBonus"] == 1.5 || prof.enemyElement == "non-but-critical" || prof.retsujitsuNoRakuen) {
    let LBCriticalArray = getLBCriticalArray(totals[key]["LB"]);
    let EXLBCriticalArray = getEXLBCriticalArray(totals[key]["EXLB"]["Critical"]);
    let normalOtherCriticalBuffArray = totals[key]["normalOtherCriticalBuff"];
    let normalOtherCriticalArray = totals[key]["normalOtherCritical"].concat(LBCriticalArray, EXLBCriticalArray, normalOtherCriticalBuffArray, totals[key]["criticalBuff"] || [], buff["criticalBuff"]);

    criticalArray = module.exports.calcCriticalArray(totals[key]["normalCritical"], totals[key]["magnaCritical"], normalOtherCriticalArray, totalSummon);
    criticalRatio = module.exports.calcCriticalRatio(criticalArray);
}
```

- those `var` also should be `let` if it's safe and possible. but at this time,
  I did not check outer scope use the variable or not. so I kept them `var` as legacy code.
- NOTE:  notice `totals[key]["criticalBuff"] || []` ... if it hard to read, then separate to another line.
  - [ ] then can remove undefined checks in calcCriticalArray (not tested other array)

----
The problem of unbalanced declaration (using `var` inside if-block)
the variable become undefined, that often make some conditional errors.
(and often make many undefined check code)

```js
if (flag) {
  var damage = 0;
} else {
  // XXX: no declare "damage" variable here
}
// "damage" is only inner scope use ?
//  or outer scope also use the variable ? ... need to search outer scope to check that.
```

- `let` can limit scope, so no need to care the variable used in outer scope.
- `var` then review will have extra task that have to check outer scope.

h-yasha:
- [x] fill empty blank tooltip
- [x] reduce clone code and make code more strict.

kei-gbf:
now last testing stage, I checked the spec.
- [x] Test Critical Buff list
  - [x] Player
    - [x] personalCriticalBuff
    - [x] CriticalBuff
  - [x] Chara
    - [x] CriticalBuff
  - [x] Test loading save data without those fields
  - [x] I do-not check conflict of buff field, that up to original critical implementation
    - because, this PR just add the buff array, no touch the critical logic. 
- [x] Add Sun-Touched Paraside (烈日の楽園).
  - [x]  Anre(water) also got Normal Might and Critical for Wind enemy

just note, 6e42d9cfbe4bd4c60561040defe531eec7850fb4 may not need now
  `totals[key]["criticalBuff"] || []` that's where undefined came from.
  when we had more unit tests, we can remove some unnecessary undefined checks safely.
  now it ok for explicit and strict checks.

----
>Implement Sun Touched Paradise in Damage Simulator.
> I would leave this for now. ..snip..

I see.
ダメージシミュレータは未対応。（現在の実装が適してない為）

PR #251, opened 2019-06-10
 Add CharaLB of ougiDamage, ougiDamageLimit, ougiGageBuff

kagekiyo7:
- Display OugiGageBuff


OrkunKocyigit:
I'll try to review these changes.
- [ ] Ougi Damage Up Calculation
- [ ] Ougi Damage Up Chart
- [ ] Ougi Damage Up Simulator
- [ ] Ougi Damage Cap Up Calculation
- [ ] Ougi Damage Cap Up Chart
- [ ] Ougi Damage Cap Up Simulator
- [ ] Ougi Gage Up Calculation
- [ ] Ougi Gage Up Chart
- [ ] Ougi Gage Up Simulator
- [ ] Interactions with support skill (Shiva)
- [ ] Save Game/Load
- [ ] Coding style

src/translate.js
```diff
@@ -377,7 +377,7 @@ var multiLangData = {
         "zh": "奥义伤害上限上升",
     },
     "チェインダメージ上限アップ": {
-        "en": "Chain Damage Cap",  //a.k.a: chainDamageLimit
+        "en": "Chain DMG Cap",  //a.k.a: chainDamageLimit
```
OrkunKocyigit: This key is same as line 370. Sets can only contain one unique key per value. Remove one of those keys.

src/global_logic.js
```diff
@@ -1792,6 +1796,9 @@ function getCharaLB(chara) {
         "ATK": 0,
         "HP": 0,
         "Element": 0,
+        "OugiDamage": 0,
```
OrkunKocyigit: Currently these three keys mapped as String objects in loop at line 1817. While it works, in order to have consistency these should be mapped into numbers as all other fields if we evaluate them as such
if (key == "ATK" || key == "HP" || key == "Element") {
                LB[key] = parseInt(chara[exactKey]);
            }

src/global_logic.js
```diff
@@ -730,6 +733,7 @@ module.exports.calcBasedOneSummon = function (summonind, prof, buff, totals) {
         coeffs["LBKonshin"] = LBKonshinCoeff;
         coeffs["other"] = otherCoeff;
         coeffs["hpRatio"] = hpCoeff;
+        coeffs["ougiGageBuff"] = ougiGageBuff - 1.0;
```
OrkunKocyigit: Removing -1.0 can be avoided by moving adding buff["ougiGage"] parameter to line 613 without any side issues.

OrkunKocyigit: Like this:
var ougiGageBuff = + totals[key]["ougiGageBuff"] + (0.01 * totals[key]["LB"]["OugiGageBuff"]) - totals[key]["ougiDebuff"];
var expectedOugiGage = ougiGageBuff + buff["ougiGage"] ;
.
.
.
coeffs["ougiGageBuff"] = ougiGageBuff; // No need - 1.0 anymore

src/translate.js
kagekiyo7: done.

src/global_logic.js
kagekiyo7: done.

src/global_logic.js
kagekiyo7: ougiGageBuff is used in line 616 & 618.
Problems occur if I change it.
https://github.com/kagekiyo7/motocal/blob/Add-CharaLB/src/global_logic.js#L616

OrkunKocyigit: Oops I miss that. The reason you choose -1 was it is default value of buff["ougiGage"]? If so is there any case where that value isn't 1?

kagekiyo7: >The reason you choose -1 was it is default value of buff["ougiGage"]? 

yes. https://github.com/kagekiyo7/motocal/blob/Add-CharaLB/src/global_logic.js#L1196

kagekiyo7: >If so is there any case where that value isn't 1?

The value 1 is to align the number of digits.
it must be coeff. (For example: 0.1 (+10%))

src/global_logic.js
```diff
@@ -1803,12 +1810,10 @@ function getCharaLB(chara) {
...
         var exactKey = "LB" + key;
         if (exactKey in chara) {
-            if (key == "ATK" || key == "HP" || key == "Element") {
-                LB[key] = parseInt(chara[exactKey]);
-            } else if (key == "DA" || key == "TA") {
+            if (key == "DA" || key == "TA") {
                 LB[key] = parseFloat(chara[exactKey]);
             } else {
-                LB[key] = chara[exactKey];
+                LB[key] = parseInt(chara[exactKey]);
```
OrkunKocyigit: This breaks critical

OrkunKocyigit: I'll check your latest commit.

OrkunKocyigit:
Latest version looks fine and tests still pass. Thanks for your hard work.

kagekiyo7:
Thank you for the review!

PR #256, opened 2019-06-13
Added Ability To Filter Combinations and Asynchronous Template Loading

OrkunKocyigit:
This PR aims to increase responsiveness of the result screen in application.

This PR contains following:
- [ ] Adds ability to filter combinations by checking grid size. global_logic.js/filterCombinations.
- [ ] Added ruleMaxSize which checks if you equipped every possible slot of your grid or not.
- [ ] Added ruleMaxSize to result.js and profile.js props.
- [ ] Passes ruleMaxSize to calcCombinations.
- [ ] calcCombinations now counts maximum possible slots in grid via maxSize.
- [ ] Added ability to toggle on/off in profile.js (Defaults: true).
image
- [x] Asynchronous data loading for arms in template.js via onDataRequested and onDataObtained methods.
- [x] Asynchronous data loading for characters in template.js via onDataRequested and onDataObtained methods.


src/translate.js
```diff
@@ -803,6 +803,11 @@ var multiLangData = {
         "ja": "チェイン数を指定してください（チェインバーストボーナスの係数が変化します）\n チェイン数1 → 0倍\n 2 → 1.25倍\n 3 → 1.33…倍\n 4 → 1.5倍",
         "zh": "设定Chain数(Chain Burst的系数会变化)",
     },
+    "弱いを隠す": {
```
kei-gbf:
```js
- 弱いを隠す
+ 弱い編成を隠す
```
編成 is "Grid" in Japanse.
- [ ] "zh" locale can be same to "en"

src/profile.js
```diff
@@ -907,6 +913,17 @@ var Profile = CreateClass({
                             </td>
                         </tr>
                     </TextWithTooltip>
+
+                    <TextWithTooltip tooltip={intl.translate("弱いを隠す", locale)} id={"tooltip-hide-grids"}>
```
kei-gbf: "弱い編成を隠す"

src/profile.js
```diff
@@ -244,7 +244,9 @@ var Profile = CreateClass({
         }
         if (e.target.type === "checkbox") {
             newState[key] = e.target.checked;
-            newState.filterOptionsChanged = true;
+            if (key.toLocaleString().toLowerCase().includes("rule")) {
```
kei-gbf: toLocaleLowerCase() ? or isn't it just match "ruleMaxSize" ?

OrkunKocyigit: I thought in the future we might add new rules so it checks "rule" keyword as a whole. I agree toLocateString is probably unnecessary and will revise it.

src/translate.js
OrkunKocyigit: Updated in 47ad220

src/profile.js
OrkunKocyigit: Updated in 47ad220

src/profile.js
OrkunKocyigit: Updated in 47ad220

kei-gbf: - .startsWith("rule")
  - (old IE does not support the method, other browsers may require polyfill)
- regex is better and flexible if the key matching rule became complex

as it already changed, "includes" is ok if no unexpected match cases.

- [x] searched "rule" may become a part of another word in wordlist
  - results are only: "rule", "overrule"  it's a rare case, currently no need to care.
  - other chance is when want to use "rule" in another compound words.
    we will need strict matching pattern, but ok not now.

kei-gbf:
:+1: for api document

- [ ] I wanted type info, is it array of number?
- [ ] Separate utility function and unit test

----
Type info

```
@param {number[]} combinations
@param {number} ...
```

I do not know about the type notation style variants,
jsdoc,esdoc, or flow ...

[ESDoc type syntax](https://esdoc.org/manual/tags.html#type-syntax)

----
- [ ] How about separate "sum" as utility function

(minimum sample)
```js
/**
 * sum([1, 2, 3, 4]) => 10
 */
const sum = (arr) => arr.reduce((total, num) => total + num);

// Hide the implementation detail in the filter logic code.
combinations.filter(combination => sum(combination) === max_size);
```

- I assume readers may not familiar to functional programming paradigm.
  - I know "reduce" function use is often a tricky for them.
    - it's being popular in JavaScript too, but it's up to their background.
  - "sum" is a popular utility function, lodash also has [sum](https://lodash.com/docs/#sum)
  - they can understand what the sum() function do without read the implementation detail.
    just write sample code in function comment, is enough.
    - I took this approach in #201 zip() function and porting other functional tools.

in JavaScript, compiler does(or may) not optimize the reduce function calls to flat loop.
if need performance tuning, we can switch the sum() implementation later.
```
function sum(arr, total=0) {
    for (let num of arr)
        total += num;
    return total;
}
```

and now sum() is reusable function, you can make unit test

- [ ] `module.exports.sum = sum;` export for tests
  - [ ] NOT: module.exports.sum = function (arr) { ...
    - Because, we can do `if run as TEST then module exports those private function`  later.
    - but no need to change others, no need wide refactoring now.

OrkunKocyigit:
> 👍 for api document
> 
> * [ ]  I wanted type info, is it array of number?
> * [ ]  Separate utility function and unit test

I implemented those changes with 7df50b9

kei-gbf:
- [ ] any cases which the filtering logic does not fit ?
  - [ ] e.g. sort by HP and HP debuff weapon ?
- [ ] Check load old save data

@OrkunKocyigit 
Can you give a sample save data which the filter works well.
I think coding status is ok. at the last, we need to test exception cases.

OrkunKocyigit:
This implementation only filters combination arrays rather than actual arm data. It is not aware of stats of the arms henceforth it won't effect anything on stat side.

Current rule only eliminates combinations that has empty slots. If I have an 6 weapons best grid will be 6 be all of them in single grid no matter what those weapons have. This makes it to motocal ignore those grids.

OrkunKocyigit:
> @OrkunKocyigit
> Can you give a sample save data which the filter works well.

You can load the application WITHOUT any save data and can see the difference.

Without:
image

With:
image

kei-gbf:
- [x] Test print out combinations length and filtered combinations length, the result were:

| Number of Different Weapons |  TotalItr | Without filterCombinations | With filterCombinations
|-- | -- | --|--|
| 10 | 1024 | 1024 | 1 |
| 11 | 2048 | 11 | 11 |
| 12 | 4096 | 66 | 66 |
| 13 | 8192 | 286 | 286 |

Seem Motocal already does the filtering when number of different weapon 11+

```js
if (isValidCombination && ((totalItr <= 1024 && num <= 10) || num === 10)) {
    combinations.push(temp);
    maxSize = Math.max(maxSize, num);
}
```

- `(totalItr <= 1024 && num <= 10) || num === 10)` 
  totalItr 1024 is when number of different weapons 10. (2**10 ?)
  -  I added weapons each count 1 ( minimum 0 maximum 1)

however, no much demerit of the filterCombinations 
I think the "Filter" has a potential to be user experience as well as "Sort"
as PR it's ok to  Approve.

- [x] Motocal make "combinations" as array now,
  the current Motocal's hard coded filtering does early filtering.
  - implement combinations as Generator + filter will be more flexible. (future topics)
    filter for generator (not Array.filter) will work better for the filterCombinations concept.
- [x] filterCombinations can optimize until 1024 iterations (10 different weapons?)


----
This is NOT change request to support those cases.
just discuss exception cases in the logic. for example, Fill the grid may not always come to the top.
Because that does not assume __debuff__ skill and __sorting key__

sorthp

kei-gbf:
@OrkunKocyigit 
I did not test in other browser,
it's using "async" keyword, we need polyfill ?

OrkunKocyigit:
Oh that is disappointing.  I don't think there is harm leaving that but I'll try to improve and add more rules in the future.
What is minimum browser versions we do aim? I am pretty sure current application without anything requires ES8 at least. If you want to aim anything lower I can add polyfill.

https://caniuse.com/#search=Async%20functions

kei-gbf:
I do not have the statistics of user-agent access log. all I know is just
./public_html/index.html said the supported version Chrome latest, iOS 10

the support status looks not bad for me.
Babel transpile them to ES5(or ES6) doesn't it?

- [ ] check Babel support and our babel configuration

I think current status is ok
if we did not have another supported browser which the code fail without polyfill,
we can't check the polyfill work or not. when anyone reported the issue we can add polyfill.

----
and I have to note, we have three ways to run js code
- main motocal code (babel + webpack)
- Storybook use different babel config. (under ./.storybook)
- Jest unit tests use Node.js engine (not browser)

----
List of new ECMA features we use now

|  | version | |
|--|--|--|--|
| let const | ES2015(ES6) | |
| Promises | ES2015(ES6) | | 
| Arrow function | ES2015(ES6) | |
| Default parameters | ES2015(ES6) | | 
| for-of loop | ES2015(ES6) | |
| Array.include | ES2016(ES7) | (not syntax) |
| async | ES2017(ES8) | |

----
https://babeljs.io/docs/en/babel-polyfill
> As of Babel 7.4.0, this package has been deprecated in favor of directly including core-js/stable (to polyfill ECMAScript features) and regenerator-runtime/runtime (needed to use transpiled generator functions):

we have now @babel/code 7.4.4

https://github.com/zloirock/core-js#babelpolyfill

OrkunKocyigit:
Hi,Yeah Babel polyfill to add support for browsers who doesn't support ES6. iOS 10 uses safari 11 which supports it no problem.On Jun 27, 2019 3:39 PM, kei-gbf <notifications@github.com> wrote:I do not have the statistics of user-agent access log. all I know is just
./public_html/index.html said the supported version Chrome latest, iOS 10
the support status looks not bad for me.
Babel transpile them to ES5(or ES6) doesn't it?
 check Babel support and our babel configuration
I think current status is ok
if we did not have another supported browser which the code fail without polyfill,
we can't check the polyfill work or not. when anyone reported the issue we can add polyfill.

and I have to note, we have three ways to run js code
main motocal code (babel + webpack)Storybook use different babel config. (under ./.storybook)Jest unit tests use Node.js engine (not browser)

List of new ECMA features we use now
|  | version | |
|--|--|--|--|
| let const | ES2015(ES6) | |
| Promises | ES2015(ES6) | |
| Arrow function | ES2015(ES6) | |
| Default parameters | ES2015(ES6) | |
| for-of loop | ES2015(ES6) | |
| Array.include | ES2016(ES7) | (not syntax) |
| async | ES2017(ES8) | |

https://babeljs.io/docs/en/babel-polyfill

As of Babel 7.4.0, this package has been deprecated in favor of directly including core-js/stable (to polyfill ECMAScript features) and regenerator-runtime/runtime (needed to use transpiled generator functions):

we have now @babel/code 7.4.4
https://github.com/zloirock/core-js#babelpolyfill

PR #258, opened 2019-06-14
BUGFIX: typo in content.js

OrkunKocyigit:
It creates some issues in some browsers like this:
image
--------
Root Cause: Typo in content.js
- [x] Changed rootRight to rootright


kei-gbf:
@OrkunKocyigit 

what browser ?
Should we search the bug detail, or just fix the typo?

- [x] style.css has .rootright
  - I suggest `root-right` naming format for those class name. (if add new names)
    css allow mix case in properties. property names are using hyphen.

I wonder why current typo class name was working correct in Chrome.
I tested chrome, it was case sensitive. is there insensitive options?

----
## Checklist 

Anyone can test those:
- [ ] Check to see the broken layout in the browser (in current repo)
- [ ] Check to see the layout fixed in the browser (in this PR repo)
- [ ] Check to see the layout in Chrome(supported browser) (in this PR repo)
  - in this case, test if wrong typo setting was working correct.
    I saw dev tool show .rootright style (typo name available for rootRight ... but don't know why)

OrkunKocyigit:
I receive that in Discord, he uses Brave Browser. I downloaded and tried but couldn't reproduce it but you can get same screen with removing rootRight from that div so I am assuming typo is the issue.

kei-gbf:
I got in chrome

- case sensitive sample: https://codepen.io/anon/pen/ewZvLP?editors=1111

```html
<!DOCTYPE html>
<html>
<head>
<style>
    .rootright {
        color: red;
    }
</style>
</head>
<body>
<div class="rootRight">TEST</div>
</body>
</html>
```

__Try remove__ `DOCTYPE` line, that make it case insensitive.

- [Case-insensitive Attribute Selector Matching Sample](https://googlechrome.github.io/samples/css-attribute-case-sensitivity/)
- [Why Are Class Names Made Lower Case?](https://support.tiny.cloud/hc/en-us/articles/226357147-Why-Are-Class-Names-Made-Lower-Case-)

----
so this patch does not effect wrong to current supported browser, the typo should be fixed.

and this mean, Motocal is now not strict HTML, no DOCTYPE html.
This will be issue when port to HTML5, we need refactor css selector names case sensitive.

OrkunKocyigit:
Cool I will create an issue about root cause. We can work on it later date

PR #262, opened 2019-06-16
Add Number of Times of Ougi

kagekiyo7:
It uses when Sort Key is related to C.A.
Some of 200% charge bar character, Kengo, Chrysaor etc.

Please give suggestions for improvement of variable, item name(eng).

I hope someone implement function to give buff only to the second C.A.


h-yasha:
I have looked into implementing this (and did) before, [only available for characters](http://yasha-motocal.herokuapp.com/).
I ended up adding duplicates of the same character for when calculating OTK-esque setups. for example Kengo would get an Unyou stack, and MH ougi effect be effective on second ougi.

(which actually remind me,) could add Djeeta to chara list (it was annoying to calculate raw ATK/etc everytime i wanted to do something) to emulate 2nd ougi better.

>I hope someone implement function to give buff only to the second C.A.

Similarly, this could be simply _emulated_ by adding more characters.

kagekiyo7:
Wow! I didn't understand about some functions, but I think it's a wonderful MOD version.

> I ended up adding duplicates of the same character for when calculating OTK-esque setups

Good idea. "Sort Key: C.A + Chain Damage" also includes Sub Character.
The problem is that copying character information is troublesome, and it is not possible to enter the master bonus of Djeeta's ATK and HP.
So maybe it would be nice to implement the function to possible to enter (and copy) second Djeeta  information in the profile field and the function to copy character information down (like a weapon tab).
(And this is also difficult for me.)

kagekiyo7:
Anyway, I will delete this pull request.

PR #265, opened 2019-06-17
Add Per Character Enemy Defense

h-yasha:
_#147 seems to be a "cleaner" way? but from what understood there were compatibility issues._

- In this branch/PR,
    - "replace" DEF Debuff row with value.
    - could also have a _Check box_ to switch between "modes" instead

- [x] enable per character enemy defense. _(yes, i'm aware about that the check box is only in 'manual mode')_
  - this could be very useful for doing OTK setups/etc...
  - I have not added this yet, since the _old_ implementation ([here](http://yasha-motocal.herokuapp.com/), check below for more details.) uses an exported _(propery?/variable?)_  in GlobalConst to "communicate" between `profile.js` and `chara.js`. which didn't seems the 'best' way to do it.

<hr>

something like this: _this is not how it is implemented in the link. (it's more "wrong" way. same concept)_
`global_const.js`
```js
module.exports.perCharaEnemyDefense = undefined; //not necessary. could be false to have "cleaner" look.
```
`profile.js`
```js
if (e.target.type === "checkbox") {
    newState[key] = e.target.checked;
    if (key == 'perCharaEnemyDefense') {
        GlobalConst.perCharaEnemyDefense = newState[key];
    }
}
```
`chara.js`
```js
{GlobalConst.perCharaEnemyDefense === true //to ignore undefined/other possible values.
 ? <tr>
    <th className="bg-primary">{intl.translate("敵防御固有値", locale)}</th>
    <td><FormControl type="number" min="1" step="0.5" value={this.state.enemyDefense}
            onBlur={this.handleOnBlur.bind(this, "enemyDefense")}
            onChange={this.handleEvent.bind(this, "enemyDefense")} />
    </td>
</tr>
 : null
}
```

### DIFF
--- a/src/chara.js
+++ b/src/chara.js
@@ -3,6 +3,7 @@ var intl = require('./translate.js');
 var {Label, Checkbox, FormControl, InputGroup, FormGroup, Button, ButtonGroup, Panel, PanelGroup, Modal, Glyphicon} = require('react-bootstrap');
 var CreateClass = require('create-react-class');
 var {RegisteredChara} = require('./template.js');
+var EnemyDefense = require('./enemy_defense.js');
 var GlobalConst = require('./global_const.js');
 
 // inject GlobalConst...
@@ -309,6 +310,9 @@ var Chara = CreateClass({
             EXLBKonshin: 0,
             EXLBDA: 0,
             EXLBTA: 0,
+            manualEnemyDefense: false,
+            enemyDefense: 10.0,
+            defenseDebuff: 0.0,
         };
     },
     componentDidMount: function () {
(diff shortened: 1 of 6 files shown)


h-yasha:
~~closed.~~ redundant with, #273. (and is much better job)

<hr>

Will change this PR to be about per character enemy defense.

kagekiyo7:
>enable per character enemy defense. (yes, i'm aware about that the check box is only in 'manual mode')

I'm not sure very well why "Enemy Innate DEF". 
Isn't "DEF Debuff" more convenient?
However, it is a very useful function.

h-yasha:
> > enable per character enemy defense. (yes, i'm aware about that the check box is only in 'manual mode')
> 
> I don't understand very well why "Enemy Innate DEF".
> Isn't "DEF Debuff" more convenient?
> (for example, OTK: Djeeta → Lyria → other chara)
> However, it is a very useful function.

ah that makes more sense. ~~why didn't i think about that before~~

h-yasha:
Will open a new PR with the proposed changes as there's much differences in master branch.

PR #266, opened 2019-06-17
Fix context in beforeEach call (unittest)

kei-gbf:
This was happen by the difference between Jasmin(CodePen) and Jest(Motocal)

It's repored in official jest issue https://github.com/facebook/jest/issues/3553
Jest fixed the re-use of `this` context as bug.
(it's actually implicitly context binding to use `this` in test code)
so Jasmin's test code will not work in Jest.

This patch is required in #201

Here is current #201 test result, it changes babel setting for jest (test only change)
jest-error

This patch only change the test code, that just moving the variable context/scope.
No effects to Motocal main code at all. So it's safe to Approve if Travis status shows green.

----

@h-yasha just mention for update unit test info
my test code not work in Jest (maybe in future when library updated)
~~we can't write ` let {variable} = this;` in Jest code anymore.~~
we should not use `this` context variable in test code.

### DIFF
--- a/src/supplemental.test.js
+++ b/src/supplemental.test.js
@@ -4,8 +4,10 @@ describe("#calcSupplementalDamage", () => {
     // generate temporary vals: [0, 0, 0, 0]
     const INITIAL_VALS = (length=4, value=0) => (new Array(length)).fill(value);
 
+    let supplementalDamageArray;
+
     beforeEach(() => {
-        this.supplementalDamageArray = {
+        supplementalDamageArray = {
             "test-buff-a": {
                 damage: 10,
                 damageWithoutCritical: 10,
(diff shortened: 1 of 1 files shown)


OrkunKocyigit:
Neither solution fails for me. How do you reproduce this error?

kei-gbf:
@OrkunKocyigit 

fail? at least current status is shown ok (TravisCI Details in this PR)

```
PASS  src/supplemental.test.js
```

The error was produced 2492295 in #201,  you can see the the full log in [Travis CI detail](https://travis-ci.org/MotocalDevelopers/motocal/builds/546346820?utm_source=github_status&utm_medium=notification)

I was missing to note that, this is patch to #201 (for now)
actually current Travis CI does not check the error case.
should I pick 2492295 to this branch ?

then I will change PR title:
"Add es2015 transform to Jest config"

The summary will be:
- this change (2492295 and this PR) will make we can write "import/export" in unit tests.
- the limitation was Jest run tests via Node.js, not browser.

OrkunKocyigit:
Oh okay, I pulled onto master and tried it. I'll try this PR changes on #201.

PR #269, opened 2019-06-21
Minor Updates

kagekiyo7:
・Add notice
・Update translate.js
Delete "Rosetta (Grand)'s Iron Maiden (5000)" because it's duration is short, and usage rate is not high.

Change enemy resistance discription to " Unite and Fight's NIGHTMARE(25%), EX+(50%)"
https://gran-matome.com/archives/57946
https://twitter.com/Granblue_GW/status/1118104526706229248

Others

・Fix supplementalDamageBuff Field
It differed from the original behavior and the entered number was reflected immediately.

・Update notice.js
・Change input-suggest placement
・Change notation in profile

---
提案：保存タブの「スキル性能・各種計算式」表削除(更新が大変なのと、スキル効果量はスキル合計で見ることができるので)

PR #270, opened 2019-06-24
Fix enemyResistance Field

kagekiyo7:
The timing at which the numbers are reflected was different.
(Processing will be slower as the number of reflections increases.)
I do not understand the mechanism, but I solved it by making it the same as others.


h-yasha:
@kagekiyo7 could you also do Supplemental Damage Buff fields?

kagekiyo7:
@h-yasha
I have already fixed it.
https://github.com/MotocalDevelopers/motocal/pull/269

PR #273, opened 2019-06-25
Allow user to manually input fields

OrkunKocyigit:
This PR aims to allow user to input acceptable manual fields without breaking current selection ability.
--------
### Code Changes
- [ ] Typeahead.js file is created. This file provides main functionality of this PR.
- [ ] Utilities.js file is created. This file provides helpful functions that might aid and assist development.
- [ ] Utilities.test.js file created for unit tests of Utilities.js
- [ ] Eslint updated to include JQuery
- [ ] TextWithTooltip class now reads placement from it is state instead of being hardcoded.
- [ ] TextWithTooltip class provides setPlacement method to update placement.
- [ ] Selectors that are used in dropdowns now have plain String or {id: any, label: String} mappers for new dropdown structures.
- [ ] Profile.js FromControl objects converted to new Typeahead objects.
- [ ] Summon.js FromControl objects converted to new Typeahead objects.
- [ ] Summon.js onHandleSummonAmount function changed. Now it doesn't require id, order etc. to be passed and work on it is state completely.
- [ ] Chara.js FormControl objects converted to new Typeahead objects.
- [ ] Component.js FormControl objects converted to new Typeahead objects.
- [ ] Typeahead.css stylesheet included in public/index.html
- [ ] TextWithTooltip now passes it is web hooks to Typeahead objects.
- [ ] Invalid style from result.js is fixed.
--------
### Profile
- [ ] Rank
- [ ] Djeeta Solo Atk Buff
- [ ] Djeeta Solo Element Buff
- [ ] Djeeta Solo Other Buff
- [ ] Djeeta Solo Other2 Buff
- [ ] Djeeta Solo Crit Buff
- [ ] Djeeta Solo DA Buff
- [ ] Djeeta Solo TA Buff
- [ ] Djeeta Solo Bonus DMG Buff
- [ ] Djeeta Solo C.A. DMG Buff
- [ ] Djeeta Solo Charge Bar Speed UP Buff
- [ ] Djeeta Solo Uplift Buff
- [ ] Djeeta Solo DMG Cap Buff
- [ ] Djeeta Solo C.A. DMG Cap Buff
- [ ] Zenith ATK Buff
- [ ] Zenith HP Buff
- [ ] Zenith DA Buff
- [ ] Zenith TA Buff
- [ ] Zenith DMG Cap Buff
- [ ] Djeeta EMP ATK
- [ ] Djeeta HP
- [ ] Djeeta Party HP
- [ ] Party Atk Buff
- [ ] Party Element Buff
- [ ] Party Other Buff
- [ ] Party Other2 Buff
- [ ] Party Crit Buff
- [ ] Party HP Buff
- [ ] Party DA Buff
- [ ] Party TA Buff
- [ ] Party C.A. DMG Buff
- [ ] Party Bonus DMG Buff
- [ ] Party Charge Bar Speed UP Buff
- [ ] Party Uplift Buff
- [ ] Party Charge Boost C.A.
- [ ] Party DMG Cap Buff
- [ ] Party C.A. DMG Cap Buff
- [ ] Enemy Innate DEF
- [ ] Player C.A. Multiplier
### Summon
- [ ] Player Summon Amount
- [ ] Friend Summon Amount 
### Chara
- [ ] Plus Bonus
- [ ] Atk Buff
- [ ] Element Buff
- [ ] Other Buff
- [ ] Other2 Buff
- [ ] Crit Buff
- [ ] DA Buff
- [ ] TA Buff
- [ ] Bonus DMG Buff
- [ ] Charge Bar Speed UP
- [ ] Uplift
- [ ] C.A. DMG
- [ ] DMG Cap
- [ ] C.A. DMG Cap
### Content
- [ ] Save Data
- [ ] Load Data


kagekiyo7:
Report:
With the smartphone UI, the list is not displayed.
image

OrkunKocyigit:
> Report:
> With the smartphone UI, the list is not displayed.
> image

Thanks I'll check and report back.

OrkunKocyigit:
Seems to work on my Android phone with Chrome. (https://ormotocal.herokuapp.com). I don't have access to iOS device right now. I'll have to check later.

kagekiyo7:
I did not find the arrow icon. (iOS11.1  Chrome)

h-yasha:
(minor) redundancy 
image

<hr>

just mention of an alternative,
_a similar solution (for a similar result), like #265. a checkbox in field header which changes the input type (or in advance one checkbox to flip them all). However, this could introduce a lot of clutter... a LOT of clutter._

OrkunKocyigit:
Okay I understand the issue. In mobile browsers browser wait until user enter any input before showing recommendations. I'll check an workaround

OrkunKocyigit:
I found a native React component that is cross platform and takes cares of all issues. I'll update once it is finishes.

OrkunKocyigit:
@kagekiyo7 Can you check it with newest version? I switch to native ReactJS component over datalist tag because how many differences each browser has. It should run perfectly now.

OrkunKocyigit:
> 
> 
> (minor) redundancy
> image
> 
> just mention of an alternative,
> _a similar solution (for a similar result), like #265. a checkbox in field header which changes the input type (or in advance one checkbox to flip them all). However, this could introduce a lot of clutter... a LOT of clutter._

You can do the same for your PR? If it satisfy your requirements you can close yours.

kagekiyo7:
@OrkunKocyigit 
Did not display... (Have you updated heroku motocal?)
image

OrkunKocyigit:
> 
> 
> @OrkunKocyigit
> Did not display... (Have you updated heroku motocal?)
> image

I did not.

OrkunKocyigit:
@kagekiyo7 
I updated https://ormotocal.herokuapp.com. Ignore "Invul Buff" button for now, I am still refactoring ^^

OrkunKocyigit:
Refactor is finished, I am pretty satisfied with current branch.

kagekiyo7:
@OrkunKocyigit 
Thank you deploy.
OK! The list is working.

----

There is a bug that whiteout occurs when you change the entered name.(PC browser also happened.)

OrkunKocyigit:
@kagekiyo7 Thanks I fixed it and deployed it.

kagekiyo7:
@OrkunKocyigit 
good. I confirmed the bug fix.

----
There is a problem that the name is not displayed in full.
image

OrkunKocyigit:
> There is a problem that the name is not displayed in full.
> image
Fixed and deployed.

kagekiyo7:
@OrkunKocyigit 
it is perfect.

----
Just a specification report(Please do what you want.):
Even if you display the list again after entering a number, the display position is not retained.

OrkunKocyigit:
@kagekiyo7 I might be misunderstanding issue. Can you provide me example if it is not a bothersome.

kagekiyo7:
@OrkunKocyigit 
Sorry for the poor explanation. It is not so difficult.
Just, the display position of the list item is reset. (The original was not done.)

kagekiyo7:
This is not a problem. It is a specification.
This is useful when re-entering numbers, but it is inconvenient when you want to add to the original numbers.

OrkunKocyigit:
@kagekiyo7 I changed focus, blur and change logics of ui elements for better user experience, you can try it on heroku.

kagekiyo7:
@OrkunKocyigit 
It works well on the iPhone.👍

OrkunKocyigit:
@kagekiyo7 It seems to work well with mouse wheel and hand gestures but scroll bar seems to be non functioning and only visual. I'll either try to make it work or remove it so it won't mislead people.

OrkunKocyigit:
@kagekiyo7 I made some optimization on smartphone.css for better scroll bar recognition. It seems working fine on chrome device emulators. Can you check it on device?

kagekiyo7:
@OrkunKocyigit 
It is such an operation.
There is a problem. Select a number to return to the original.
https://i.imgur.com/ViukHFq.mp4
(Since the resolution is high, please reduce the screen.)

OrkunKocyigit:
LOL it only happens in chrome. I dunno what causing it but I'll try to figure it out.

OrkunKocyigit:
@kagekiyo7 Okay this bug happens in Chrome if any child container inside of the go overs the main div like in that defense type list. I changed css so it won't but will add scroll bars to root div instead and it fixed the issue.

kagekiyo7:
@OrkunKocyigit 
It is such an operation. The position is returned to the center, but it does not look so beautiful.
https://imgur.com/a/9SNj5jB

By the way, it is like this in the original motocal.
https://imgur.com/a/ATJHGEs

----
I have two plans. (I'm not a programmer, so for your reference)
- Only the smartphone UI can not be input directly. (Original UI)

On the iPhone, the keyboard window comes out, which is an obstacle to selection.

- Fix the box being input completely in the center. (Position where all items in the select box can be seen)

You won't be bothered by the scroll bar problem.

OrkunKocyigit:
@kagekiyo7 This happens only when you scroll up or down and menu goes out of focus right?

OrkunKocyigit:
@kagekiyo7 
Bad news: It wasn't a simple css issue.
Good news: I fixed it and deployed into heroku.

kagekiyo7:
@OrkunKocyigit 
👍 There are no scroll bar restrictions. (I misunderstood that you were going intentionally.)
https://imgur.com/a/XXGQkIH

----
Can you reverse the order of the master bonus select boxes? (To make it easy to choose the maximum value)
image

----
Just report(Not a change request):
In the original motocal, the display direction was considered so that the select box was not displayed off screen.
imageimage

OrkunKocyigit:
@kagekiyo7 I added flip functionality and reversed master boxes.

kagekiyo7:
@OrkunKocyigit 
Thank you! Can you put the order of layers in the select box up?
image

OrkunKocyigit:
@kagekiyo7 I implemented and deployed to heroku.

kei-gbf:
@OrkunKocyigit 

sorry I'm late to notice your request,
I had idea for the implementations

https://codepen.io/anon/pen/zVRqBJ?editors=0011

- auto complete sometime have unexpected behavior (I had tried he bottom of #147)
  - Combobox is also one of choices
  - look the codepen sample, it will be easy to switch the form component
- library dependency
  - global_const.js import the library is ...
    - the code use no use consts in global_const? then they can be new file (e.g typeahead.js)
    - global_const does not need the library dependency
    - if just want to import via global_const then
       - implement the library specific code in typeahead.js
       - global_const import the typeahead.js and re-exports
       - __readers can notice where typeahead specific codes__

I have no much time to explain, now.
this is just the first impression, I did not look all changes yet.

in the codepen demo, I aimed the feature without change the current code.
but I am not sure about event handlers. for now just share the concept and idea.

----
Now you have edit a lot FormControl component
I introduce use of storybook for view tests (if it was not late)

1. implement the new feature as React component
  - because, hard coding change make side effects widely
     a lot modify event handlers make the wide dependency
     that's usually going to be hard to maintenance when the code is grown.
2. add the component to storybook
  - while in test/debug stage, no need to make same changes to all others until finish test pass
  - storybook can be the playground to test the UI
3. after pass all requirement
  - integrate to Motocal code
  - XXX: but event handlers ... React's event system is single dispatch
    then you had to edit current handlers (that mix dependency code and legacy code)
    I do not know the details yet. maybe not so easy as I am thinking.
    if multiple dispatch was supported, you could separate those events.

OrkunKocyigit:
It is fine, I won't be able to work on it until I finish my 20 boxes in event. I will check on it later.On Jun 30, 2019 3:52 AM, kei-gbf <notifications@github.com> wrote:@OrkunKocyigit
sorry I'm late to notice your request,
I had idea for the implementations
https://codepen.io/anon/pen/zVRqBJ?editors=0011
auto complete sometime have unexpected behavior (I had tried he bottom of #147)
Combobox is also one of choiceslook the codepen sample, it will be easy to switch the form component
library dependency
global_const.js import the library is ...
the code use no use consts in global_const? then they can be new file (e.g typeahead.js)global_const does not need the library dependencyif just want to import via global_const then
implement the library specific code in typeahead.jsglobal_const import the typeahead.js and re-exportsreaders can notice where typeahead specific codes



I have no much time to explain, now.
this is just the first impression, I did not look all changes yet.
in the codepen demo, I aimed the feature without change the current code.
but I am not sure about event handlers. for now just share the concept and idea.

Now you have edit a lot FormControl component
I introduce use of storybook for view tests (if it was not late)
implement the new feature as React component
because, hard coding change make side effects widely
a lot modify event handlers make the wide dependency
that's usually going to be hard to maintenance when the code is grown.
add the component to storybook
while in test/debug stage, no need to make same changes to all others until finish test passstorybook can be the playground to test the UI
after pass all requirement
integrate to Motocal codeXXX: but event handlers ... React's event system is single dispatch
then you had to edit current handlers (that mix dependency code and legacy code)
I do not know the details yet. maybe not so easy as I am thinking.
if multiple dispatch was supported, you could separate those events.

kagekiyo7:
@OrkunKocyigit 
I'm sorry for my poor explanation. Is this possible?
image

OrkunKocyigit:
By hidden select box you mean you don't want it to pass tabs? It is possible to do that by giving high zindex to tabs you want to be on top

kagekiyo7:
@OrkunKocyigit 
Yes. In the original motocal it was. 
I think that is better.
image

OrkunKocyigit:
@kagekiyo7 How did you create it? When I try it goes under it as it should.

image

kagekiyo7:
@OrkunKocyigit 
The original motocal I say is master.
https://motocal-groovman.netlify.com
So this is a bootstrap select box

kei-gbf:
@OrkunKocyigit 

Yes, that approach look good for me.

I think Ideally, it's possible to switch the component without any change current JSX code.
Implement the custom component as adapter to pass original event handler.

one of the big merit is, make it easy to switch to another library.
because library dependency code are all within the separated file.

More demo:
https://codepen.io/anon/pen/EBQGKZ?editors=0111

- issue: current code pass this.props.children and event handler, requires adapters
    - need convert this.props.children  fit for the component
       in the component side
    - adapter for original event handler

and the adapter code is the library specific parts.

with storybook, it's possible story to switch current FormControl component
"withKnob" like change the theme. you can test without touch current working code,
this approach won't not change "logic" of the current working code.
that can reduce test/review costs a lot. (if widely changes, it take long time for review/tests)

kei-gbf:
@OrkunKocyigit 

I made the demo branch (storybook and switch component)
you can try:

1. Open GitPod [add-swith-component-storybook](https://gitpod.io/#https://github.com/kei-gbf/motocal/tree/add-switch-component-storybook)
2. `npm --save install react-widgets`
  2.1. (I just forgot to add in commit, note spell is not "widget", it's "widgets")
3.  `npm run storybook`
4.  (wait a compile then) Expose -> Open brower 


switch-storybook

- Player/Chara/Summon/Weapon -> Knob -> Backend
  - you can test switch FormControl in actual view
    - KNOWN PROBLEM: in this demo, a lot field replace unexpected. (won't fix)
- Components -> Switch form control
  - "Knob" tab to switch the backend
  - "Action" to debug the event handler works
    - KNOWN PROBLEM: Currently, combobox ignores onBlur (won't fix)

the component source in storybook is:
```jsx
<SwitchComponent componentClass="select"
    debug={true}
    backend={backendSelector('form-components-backend')}
    onChange={action('onChange')}
    onBlur={action('onBlur')}>
    {[1,2,3,4,5,6,7,8,9].map(value =>
        <option key={value} value={value}>{value}</option>)}
</SwitchComponent>
```

SwitchComponent is the wrapper component.
in actual code, assume the component name FormControl.

you can add component here, like unit tests. e.g:

- min max implementation for the select/manual input
- type="number" component ... to check current existing component work via wrapper.
- other select component which does not need manual input

----
- NOTICE: profile/chara/summon/armlist has only few changes
  to be switch-able.

```js
// import
const {getFormControlClass} = require('./form-components');

// in render function (wraps FormControl)
var FormControl = getFormControlClass(this.props.backend);
```

----
(won't fix) Because, This demo just show the concept and workflow of use storybook.
I noted hints in source comment, but I did not test them now.

In this demo, I want to introduce:

- [ ] Switch component concept
- [ ] use of Storybook for UI improvement

OrkunKocyigit:
@kagekiyo7 I containerized logic. It also includes function you requested.
image

kagekiyo7:
@OrkunKocyigit 
👍

Can you change the order of layers?
selectbox > tooltip
image

----
Report:
In the original motocal, the scroll bar could not be moved if the select box was open.
However, with this pull request, the scroll bar can be moved, which causes display problems.
There is no problem in use, so there is no need to deal with it.
image

OrkunKocyigit:
> @OrkunKocyigit
> 👍
> 
> Can you change the order of layers?
> selectbox > tooltip
> image

I think that may be bad way to implement as it will dissappear when dropbox flips. Better way to handle would be flipping textbox to opposite location of dropbox. (Like in badly photoshopped pictures below):

Dropdown, Hint Up
image

--------

Dropup, Hint Down
image




I'll check Hintbox and see if we can edit it.

OrkunKocyigit:
@kei-gbf I am working on refactoring scripts and writing unit tests for non-ui methods. I'll try to check storybook branch when I am finished with that.

OrkunKocyigit:
@kagekiyo7 I implemeneted and pushed function in my comment. It should auto deploy in about 5 minutes.

kei-gbf:
@OrkunKocyigit 

I put #282 as Draft PR (but will not merge, just for demo)
to not mess main topic here.

kagekiyo7:
@OrkunKocyigit 
It seems to me working well.👌
The operation on the iPhone is like this.
https://imgur.com/a/AEwAzKy

----
It is not%. Real number.
image

OrkunKocyigit:
@kagekiyo7 It opens numeric keyboard right? I dunno enough japanese to tell it.

if you type the option and press enter, it will select correct option without tapping on it.

kagekiyo7:
>if you type the option and press enter, it will select correct option without tapping on it.

option? I’m sorry I don’t quite know, but we can switch or close the keyboard.
The reason why I said it's hindrance is because both the keyboard and the select box appear.
Anyway, this is probably a less problematic issue.

OrkunKocyigit:
@kagekiyo7 Fixed.

OrkunKocyigit:
@kei-gbf I would say current branch satisfy your requirements and can be considered for review.

This PR requires npm update.

kei-gbf:
### Check Specification

- [ ] Check those 55 component really need manual input?
  - [x] FormControl componentClass="select" ... ok
  - [ ] FormControl type="number", why need to replace them?
        it already allow manual input and min/max check.
- [ ] Save/Load ... check the component can store loaded initial value.


### Code design

- [ ] there are a lot changes in profile/chara/armlist/summon files.
      that make props fit for the Typeahead component.
      those widely changes require review to check all component code, and they all similar things.
- I show in my demo, using Adapter pattern then reduce coding review costs.
      (but have to tests all those components work, that's view integration test, not code review)
- SOLUTION: with Adapter pattern, convert those interface gaps within the Typeahead component side,
  or create the component via Adapter
  - provide wrapper which replae current "FormComponent" (then no need change JSX code at all)
    this will reduce much review costs.


### Coding review

- [ ] "validateNumber" has many issues
- [ ] more ...

I did not check all, so I will not report this yet, 
we may not need some of them by change the approach.


- :+1: it's good to have API doc and Unit tests.

however, other early refactoring may be worst time,
you had works a lot, but if no need most of them, it's unnecessary optimization.


I could not review this PR early stage, becauze we have many PR now.

OrkunKocyigit:
@kei-gbf I think how I know adapter pattern does differ from your definition in your PR.

What I know "adapter pattern" is a middleware class that merges different types of data structure such as different REST API in composite manner in order to show in a single component.

For example if we decide to show charadata.json and armdata.json into same CustomComponent, we may use adapter pattern in order to make those two compatible as they have different responses.

But from what I understand, your use case is converting one jsx dom into another dom using an adapter? If so I don't see any benefit of such use case.

kei-gbf:
@OrkunKocyigit 

It's actually, different to the definition, I should not use "pattern"
I want reuse here is existing the JSX code for FormControl.

----
it's a point reduce the range of change side effect to minimum.

- [ ] keep current FormControl JSX code (I explained the merit of this)
  - not performance benefit, it's reduce cost/risk of changes a lot code.
    convert runtime is not important,
    but it's not every rendering time. constructor can cache the converted data.

This benefit is no change current JSX,then no need to check 55 component jsx codes. (need only view tests)
those changes in chara/profile/armlist/summon are the custom component specific,
I think event handlers are improved, they are now in the component,
the convert props is still issue.

The original code has the issue, a lot convert for <option>
that should do in component side, then no need many similar code,
but now my focus is on the range of changes code.

no need this optimization now (because "diff" will show wide changes)
I want separate PR for that optimization tasks. (addon props too)

>  FormControl type="number", why need to replace them?
it already allow manual input and min/max check.

at this point, I misunderstood. it was not replace all type="number" components.
I wanted separate coding format topic to another PR or at least another commit.
this make extra "diff", reviewer does not know it change props or just format.

it's ok for small changes, but this PR has a lot component changes.

## UI/UX

- [ ] Check the autocomplete
   I had met sometime unexpected behavior
- [ ] Check for options label != value, e.g. enemyDepence

----
I expect finally, the review cost would be simple
- [ ] review the custom Typeahead component (and utilities)
- [ ] test all view works

below can be another topic, or at least after pass reviewed once.
then standalone commit will show the independence diff.
those changes is ok but I want take a step for smooth review.

- [ ] addons
- [ ] reformat existing jsx code

kei-gbf:
## Code review

```js
const validateNumber = function (selected, min, max, previous) {
    if (isNaN(selected)) {
        if (previous) {
            if (!isNaN(parseFloat(previous))) {
                return parseFloat(previous);
            } else {
                return 0;
            }
        } else {
            if (min) {
                return min;
            } else {
                return 0;
            }
        }
    } else {
        if (selected < min) {
            return min;
        } else if (selected > max) {
            return max;
        } else {
            return selected;
        }
    }
};
```

- [ ] many branches -> require a lot tests
- [ ] double-call parseFloat
- [ ] if isNaN(selected)
  - [ ] no max check ... (just see as variable "max" is not used in the block)
  - [ ] no min/max check for previous ... no problem in this case?
  - [ ] default is min or 0 up to the condition
    - when reading code, readers can't get what the default value
- [x] if selected is number ... ok
  - isNaN can pass string type object which has number expression
    - [ ] isNaN("Infinity") -> false, because JavaScript implicitly type conversion.
  - selected="Infinity" (string) ... maybe no need to care such Irregular cases
- validateNumber is a private use? or other module (except unit test) use the function?
  - if private use only then you can avoid those illegal case by limit the usage of the function.


To simplify this validation, you can separate function roles

- [ ] min/max check ... like lodash "clump" Equivalent, or "inRange" if just check
- [ ] type chack
- [ ] default value


```js
function clamp(min, max, num) {
   return Math.min(max, Math.max(min, num));
}
```

simple function is more reusable as utility.

- [ ] see also popular implementation
   - lodash clamp (the popular utility library has clamp function to return value within min/max
   - just order of arguments are different point, either has merits
     - low middle high -> looks easy to understand for code reading
     - min, max, num -> this order useful for bind parameters
       - use-case: this.validator = clamp.bind(null, props.min, props.max);


----

it's an unbalance, getValidatData handle treat default value for "text",
getValidData can do that task for "number" too.

```js
const getValidData = function (type, min, max, selected, previous) {
    switch (type) {
        case "number":
            return validateNumber(parseFloat(selected), min, max, previous);
        case "text":
            if (selected) {
                return selected.toString();
            } else if (previous) {
                return previous.toString();
            } else {
                return "";
            }
        default:
            return selected;
    }
};
````

-  case "text": `return (selected || previous || "").toString();` for simplify
  - note, but not for number, this can't check 0 or undefined
- can pass parseFloat(previous) then no need parseFloat twice in validateNumber()
- [ ] no need to be one mixed unction, "text" does not need min/max parameters
    - Component can have different validator (separate implementation numberValidator and textValidator)


as "validator" role, they usually just check value is valid or not.
"return the default value", may be implicit, users may confuse. (I did not test this yet)

File: src/utilities.js [hunk 1]
```diff
@@ -69,68 +53,79 @@ const filterObjectsFromSave = function (save, list) {
...
+    if (!isNumeric(num)) {
+        if (!isNumeric(defaultValue)) {
+            num = 0;
         } else {
-            if (min !== undefined) {
-                return min;
-            } else {
-                return 0;
-            }
+            return getValidNumber(defaultValue, min, max);
```
kei-gbf: no need recursive call
`num = defaultValue;`

src/utilities.js
```diff
@@ -69,68 +53,79 @@ const filterObjectsFromSave = function (save, list) {
...
+/**
+ * Returns a valid text
+ * @param val Value to be check
+ * @param defaultValue Default value to be passed
+ * @returns {*} Valid String
+ */
+const getValidText = function (val, defaultValue = "") {
+    if (val === undefined) {
+        if (defaultValue === undefined) {
+            val = defaultValue;
```
kei-gbf: I don't get here, `val = undefined` ?

src/utilities.js
```diff
@@ -69,68 +53,79 @@ const filterObjectsFromSave = function (save, list) {
...
         } else {
-            return selected;
+            return getValidText(defaultValue);
         }
+    } else if (typeof val !== "string") {
+        val = val.toString();
     }
+    return val;
```
kei-gbf: try `function getValidText(val, defaultValue="") { return (val.toString() || defaultValue); }`
after made enough unit test, this simple code may satisfy the requirements.

src/utilities.js
```diff
@@ -69,68 +53,79 @@ const filterObjectsFromSave = function (save, list) {
...
-                return 0;
-            }
+const clampNumber = function (selected, min, max) {
+    if (selected < min) {
+        return min;
+    } else if (selected > max) {
+        return max;
+    } else {
+        return selected;
+    }
```
kei-gbf: eslint may warn the else-return

```js
if (selected < min) {
   return min;
}
if (selected > max) {
    return max;
}
return selected;
```

- no "else if" chain, it's ok for this simple case and easy to move the code block.
  - change max-check block first, if it's used more than min-check. (very small optimization)
    it has extra format task if modify "else if" chain.
  - but result may change by which min or max first, when invalid parameter given like "min > max"
- (just info) single return is better sometime, if possible. I explain merit of lodash clamp approach

```js
if (num < min) {
    num = min;
}
if (num > max) {
    num = max;
}
// can be break-point for debug, all cases pass here.
// print debug too, if three return statements, then logging points also three.
return num;
```

no need to be too strict in this simple function.

kei-gbf: it was up to the option: https://eslint.org/docs/rules/no-else-return

src/utilities.js
OrkunKocyigit: if (string) will produce false as a result if string is empty. Empty string can be considered valid text henceforth I check if string is defined or not

src/utilities.js
OrkunKocyigit: I added this in case default value provided to function is not a valid number. Replacing it fails following unit test:
expect(Utilities.getValidNumber("ABC",undefined,undefined, "ABC")).toBe(0);

src/utilities.js
OrkunKocyigit: Oh I replaced that code block back then. I am comparing now... It should be fixed

src/utilities.js
OrkunKocyigit: Following unit test fails with this code:
expect(Utilities.getValidText(undefined, undefined)).toBe("");

src/utilities.js
OrkunKocyigit: Fixed in 8902cef

OrkunKocyigit: Lodash approach doesn't work if any of the min, max values are undefined which is a valid use case.

kei-gbf: this is just show merit of single return.
undefined check is not responsibility of this function, callee should check
and in use case max or min undefine, should use Math.min/max, clamp for both check at once.

src/utilities.js
kei-gbf: ok I wonder that could not find in a head, it old outdated commit

kei-gbf:
I will write out tasks for review
I feel we are not sharing this details:

- [ ] check a lot component jsx code
  - [ ] it's now mixing many topics, that make diff complex. -> review take a time as number of components
    - those change are costs like copy&paste clone code
    - do within component side, can reduce this cost
       then code review is only the component,
      no need to check those widely jsx changes.
  - [x]  change FormControl -> Typeahead ... this is ok, and necessary change.
  - [ ] addon ... this is an extra task
   - [ ] I want them separate commit
  - [ ] and another "options" list convert etc too. it's Typeahead specific changes
    - [ ] convert the list in component side
      - [ ] if you want optimize the `<option>` do it later.
  - [x] keep previous event handler ... ok
     - [ ] but not expect the existing event change for Typeahead, this make extra cost
            I mean on this review stage, (I see the "diff" here, and the change costs/risks)
  - [ ] coding style format ... this take the widely changes, 
    the diff hide important change which should be reviewed.
    remove this change can save a lot review costs for jsx changes.
    the format is ok for small topic, but for this topic (a lot components) it much distrub diff.

it not just review costs, merit is that can keep current FormControl compatibility
and easy to scale to another library, by capsulize Typeahead specific code into the component.

src/utilities.js
kei-gbf: no need "selected" variable name, as general utility function.
not selected component specific code now.

OrkunKocyigit:
> coding style format ... this take the widely changes,
> the diff hide important change which should be reviewed.
> remove this change can save a lot review costs for jsx changes.
> the format is ok for small topic, but for this topic (a lot components) it much distrub diff.

I am assuming you are talking about this right?

let showInvul = this.state.openBufflist ? "" : "hidden";

I added that because current motocal lags a lot when you press that button. It is a lot expensive for browser to create and add/remove from/to the whole dom then just set css to display:none to display:block. It is not directly related to my PR but more of a UX thing. I can remove that if it causes problem for you.

kei-gbf:
@OrkunKocyigit 

no, it's about JSX code in profile/chara/armlist/summon
a lot <FormControl components changes, I expect the diff only FormControl -> Typeahead
component rename. but the format change mess, then code review have to check props changes etc
those are unnecessary extra tasks for review. (addon props is acceptable but not on this stage)

OrkunKocyigit:
Oh I fixed that uneven divs that was causing problems with odin summon. It is more of a unrelated bug fix than being related to Typeahead. You can see difference between addon buttons.

Master:
image

My Branch:
image

src/utilities.js
kei-gbf: I should check how the function was used,
defaultValue need check undefined? if it was runtime value,
but then  no min/max checks for defaultValue.
what does the function do for min/max undefined case?

and still no need recursive call, the review have to check it not go infinite loop or
that take an unnecessary risk. flat code is better.

OrkunKocyigit:
Okay I am trying to understand issue here. My current understanding is that you don't want **ANY** jsx changes other than FormControl -> Typeahead

Current JSX:
<FormControl  componentClass="select" 
			 value={this.state.normalBuff}
			 onChange={this.handleSelectEvent.bind(this, "normalBuff")}>
			 {selector.buffLevel}
</FormControl>

Should work as this or not work at all.
<Typeahead     componentClass="select" 
			 value={this.state.normalBuff}
			 onChange={this.handleSelectEvent.bind(this, "normalBuff")}>
			 {selector.buffLevel}
</Typeahead>

kei-gbf:
@OrkunKocyigit 
current master has the issue ? I can't test right now.
I also met the same problem in my demo branch, then fixed by passing className.

if motocal/master had the issue, then can you make another PR for the fix?

----
>  is that you don't want ANY jsx changes other than FormControl -> Typeahead

Yes, that can reduce really much review time/cost
check 55+ more jsx code -> just component name and other stuff inside component.
(diff has ignore whitespaces option for ignore such format only change, but still a lot diff)

and update my previous comment, event handler too I want revert them once.
it's ok to add another improvement (addon) after pass the component changes review.

OrkunKocyigit:
I'll close this PR. I am not interested removing all the additional UI/UX functionalities I added in this PR for testing sake of it.

kei-gbf:
Ok, summary of reason that reject this PR

- Change of legacy code which take unnecessary risks/costs.
- change event handler
  - effect to existing components which no need to test
  - currently, we don't have tests for event handlers, manual tests is costs.
- convert option list
  - require many check which does not need if convert once in component side.
  - legacy code has an issue, we can improve it later. but expect it separate commit.
- addon props
- coding style format
  - mess for diff (in this situation, mix many topics and a lot changes)

----
This PR changed a lot components.
the effect is wide, diff contains unnecessary changes, that make code review hard.
it's possible to implement the PR feature without such wide side effects,
that will be more easy for test/review.
I've shown the approach #282 demo, but I will not do that task.

This will close #282

kagekiyo7:
@OrkunKocyigit 
I'm sorry. I regret that I did the operation test first.

OrkunKocyigit:
@kagekiyo7 It is fine. I understand @kei-gbf concerns about modifying legacy code. Most of the functionality I provide in this PR depends on prop values for layer abstraction. Without it, it is just shoddy select box. I'll work on it on my git.

kei-gbf:
@OrkunKocyigit , @kagekiyo7 

it does not mean reject the PR feature, I was going to suggest  is improve the process.
you are doing at the first of this PR, are View tests.
those improvements are ok. it's not useless.

I expected the process is, each commits had independent topic then easy to check the diff.
but as I saw logs, not easy to revert specific codes for this case.
it's better to do in new PR with tidy commits.

- [ ] commit: add custom component and rename component name only
- [ ] commit: addon props
- [ ] commit: improve options list
-  not change legacy event handler. should inside the custom component event if possible.
   because of that will/may effects to another components.
- min/max checks and so on, other enhancements.

@OrkunKocyigit 
I am sensitive for changes of the legacy code, because of in this case it's a lot components.
if it was few, then not so problem, it's easy to check manually.

I have few more topics about another utilities function.
I will just leave note as info, later. (coding topic, it's minor thing)

----
55以上のコンポーネントのコードレビューが発生するのですが、
現状のdiff には、インデント等の修正や、プロパティ等の変更等が混在していて、
diff での変更点が把握しずらく、レビューやテストに掛かる時間も大きくなります。
(数が少なければそれ程問題ありません)

- コンポーネント名の変更
- addonプロパティの導入
- コンポーネントに渡すリストの最適化

がそれぞれの独立したコミットになっていると、差分が把握しやすくなるので
手順を追って変更するように提案したのと、

- 既存のイベントハンドラの変更
  影響範囲がさらに広くなる為、変更要求したかった部分。
  (詳しく追ってませんが)影響がある場合、他の既存コンポーネントの再テストも発生します。


変更の重複を省き、コミットを整理することで
コンポネントの追加部分のコードレビュー（コード重複なし1か所）と、
新規で変更するコンポーネントのビューのテスト（50数か所）のテストのみで良くなります。
（最後のテストは現状、自動化テストがない為、避けられない箇所）

PR #277, opened 2019-06-28
4* Taisai Spirit Bow

Groovman:


h-yasha:
LGTM!

kagekiyo7:
Commit is fine.
However, the commit history will be duplicated.

Groovman:
Yeah we should not squash :)

PR #282, opened 2019-07-02
[Demo] Switchable component and Storybook

kei-gbf:
__This PR is just a place for demo, will not be merged because there are uncompleted tasks__

ref: #273

This branch just show use case of Storybook
for UI testing, and Switch component class.

Points of this demo

- no much changes for current codes profile/chara/summon/armlist
  - reduce review/test costs
- dependency are in separated file
- adapter layer will handle the defference of interfaces.
- use storybook for the playground, debug does not mess main code.

switch-storybook

----
There are a lot uncompleted tasks, I will not touch them in this branch.

- naming of component/library name (a bit conflict/confusable)
- implement more events/props adapters
  - onBlur
  - min/max
- unexpected replace
  - this demo replace ALL select component
    but no need manual input for non-numeric value.

----
NOTE: this is NOT suggesting react-widgets library, that will gain library size.
I used them just for demo.


kei-gbf:
## Components

c1

- [ ] Check the code diff of "profile.js", "chara.js", "profile.js", "armlist.js"
     only few changes, no touch current code.
- [ ] switch the backend component and adapter handle the difference of interfaces.
     - This demo won't complete all tasks
     - those backend adapter are different variable name in source code.
- [ ] easy to test another library in this design
- [ ] reduce test/review cost, this approach does not add clone code.

----
c2

- [x] library dependency within a custom component :+1:
- [ ] there are still the custom component dependency changes widely
    - [ ] same changes for each components, make clone code -> increase tests costs

kei-gbf:
## Workflow for UI Improvement

- Implement new feature as React component
- Add story to storybook (like a unit tests for view)
  - explain use cases would be useful catalog
  - in this situation (SwichComponent) view need to tests
    - load non-select component via this adapter (type="number" case)
      - if any state flag then make it select-able by "Knob" (storybook plugin)
    - event handler test in "Action" tab
      - the console display values passed via the event object. (useful for debug)
   - [ ] can be document catalog for component
         I do not setup that environment yet, there is plugin for esdoc.
- integration to main code, view is still testable via Chara/Arm/Summon/Profile
  - [ ] currently, missing result page, haisui graph etc ... it's work in progress.

### what storybook can improve
- debug of clone code, testers have to check widely similar things to many fields
  - it happen sometime "field A" had a problem, but "field B" is ok
   because by missing manually copy&pasted change.
   no need to test widely in this stage.
   at the first stage, test single component is ok
- this save develop time too

kei-gbf:
This branch got the report "All checks have failed"
(won't fix in this PR) unit test was passed in local v10.15.3
not main topic of this demo.

- [ ] polyfill Array.prototype.values? but the most browser support the method now.
      it's jest/node side's problem.
- [ ] update node version, or babel config for jest

kei-gbf:
ref #273 was cloed

PR #287, opened 2019-07-05
Fix Each child in an array or iterator should have a unique "key" prop

OrkunKocyigit:
More info at https://github.com/OrkunKocyigit/motocal/issues/11


kei-gbf:
- [ ] `key={"remainHP"}` is equal to `key="remainHP"` ?
  - __EDIT__: I noticed, that following id's style. it's minor stuff not effect to topic.

PR #290, opened 2019-07-05
 Organize information displayed at the top

kagekiyo7:
Is it better to delete "Buff for Party:" and indent instead?

master
image

this
image


h-yasha:
I think "Buff for Party" is better. more clarity that these are applied to the party.

kagekiyo7:
Understood. Thank you for your opinion.

PR #294, opened 2019-07-09
Limit expectedOugiGage to min 0

h-yasha:
- It is possible to get negative charge bar raise per turn (through support ability, applying negative charge bar buff in party buff and indv buff). This PR limits it to be 0 and above. no negative values.

- current interaction: 
image
- adding a C.A. DMG Buff 
image

- with this PR:
image


kagekiyo7:
In granblue, there is a rounding up of the decimal point.
So 1.0 will be guaranteed.

`expectedOugiGage = Math.max(1.0, expectedOugiGage);`


Since motocal calculates expected values, is this not necessary? I do not know.

h-yasha:
> In granblue, there is a rounding up of the decimal point.
> So 1.0 will be guaranteed.
> 
> `expectedOugiGage = Math.max(1.0, expectedOugiGage);`
> 
> Since motocal calculates expected values, is this not necessary? I do not know.

it's expected ougi gauge **per turn**. so, yes, 1.0 will be guaranteed. 
changed it to 0 negative values only. and "round up" any value between 1 and 0.

PR #305, opened 2019-07-16
Fix whiteout in result.js

kagekiyo7:
When I opened the existing save data and display skill amount, motocal occurs white-out screen.
It seems that ougiRatio or ougiFixedDamage or ougiBonusPlainDamage is not recognized as a number.

### DIFF
--- a/src/result.js
+++ b/src/result.js
@@ -1379,7 +1379,7 @@ var Result = CreateClass({
                                     <span key={key + "-" + skillKey}>
                                             <span
                                                 className={"label label-" + labelType}>{intl.translate(label, locale)}</span>&nbsp;
-                                        {(m.data[key][skillKey]).toFixed(1)}&nbsp;
+                                        {parseFloat(m.data[key][skillKey]).toFixed(1)}&nbsp;
                                         </span>
                                 );
                             }


OrkunKocyigit:
It should be responsibility of variable itself to be float if it is that way rather than generic wrapper for all variables. Generic solution may cause problems in future when we wanna add different types of variables or if it is defined. Can you check why it becomes non float instead?

Groovman:
@OrkunKocyigit 
React input fields have some problems with spitting out floats, they give out strings sometimes instead of float and you end up with concated string instead of a addition when you use a + instruction. Multiplication is save when it spits out a string.

kei-gbf:
it's save data issue ... missing initialize ?

- master repo have the problem, need hot fix soon (it should be high priority)
- remove the cause
  - whiteout, may not only the display problem. when it has unexpected value
    that may have effects to calculation result etc. at least we need to check.

parseFloat is for type conversion,
e.g.  parseFloat(undefined) => NaN. at least avoid whiteout, but that will show NaN
if it was undefined check then (m.data[key][skillKey] || 0).toFixed(0) can be quick dirty fix.

- [x]  I test load save-data which remove "ougiBonusPlainDamage" field
- [x] no exists "ougiFixedDamage" in save data currently. 
- [ ] error when remove ougiRatio field
  - when ougiRatio was added, there was no show result

ougiRatio
the cause is seem, default value for ougiRatio is only used in ougi calc function,
but the problem is the value was not passed to result.

kagekiyo7:
Thank you everyone for your opinion.
I created PR. Is there a problem with this?
https://github.com/MotocalDevelopers/motocal/pull/307

kei-gbf:
@OrkunKocyigit 
I agree the point responsibility of variable itself.
we can continue to discuss in #307 

as quick temporary fix, is this PR patch ok ? the status is: as I test
at least it works to avoid whiteout but show NaN for old save data.

for generics, currently only three fields #288 
ougiRatio, ougiFixedDamage, ougiBonusPlainDamage
IMO plain and fixed damage should be Integer, ratio only Float ... toFixed also forced the type float.

I suggest the two steps solutions, I Approve now please Merge if it's ok

- 305 hot fix by parseFloat (take the priority)
- 307 resolve the cause, and revert the temporary parseFloat added in this PR

TODO: pls add the #288 to PR comment as ref

PR #308, opened 2019-07-17
World of Death and Love (Nier) field effect

Groovman:
Added World of Death and Love field effect.
Moved arcanum evoker field effects to Party Buffs

Needs: JP/CHN translation.


kei-gbf:
> Needs: JP/CHN translation.

I found Chinese wiki yesterday but I couldn't read at all
There are two types of locale for them, it may not good to mix for them.
zh-HK or zh-CN as read translate.js Motocal use zh-CN (translate function)
but I did not know the Chinese wiki's localce hk or ch.

for now, fill "zh" as "en" text is ok. until get active translation maintainer.
"zh" and "ja" we share some Kanji letters but not for the most cases.

I think translate function should return "en" as default "zh" case.

kei-gbf:
I have a question/interest about the 20% __unique__

Maria Theresa 's 20% damage up (support 3) is reported same field with Shield of Eternal splendor.
I want to know verification if other Sages also use the same fields. (maybe Kaimu too)

Motocal does not implement the Wonder's effects? this info will not bother this PR.

kagekiyo7:
Is additionalDamage dark element?
(Since there is no such function in motocal now, it is TODO.)

IMO:
Duration is 3T, Cooldown is 6T. 
I think this can not be used other than only 1T calculation…

- jp translation
`"ja": "パーティ全体に別枠20%、追撃30%を付与 (ニーア)",`

>CHN translation

IMO:
I do not know the reason of being of Chinese translation.
Because, Granblue Fantasy supports only Japanese and English.
I think Chinese likely understands either language.

Groovman:
@kagekiyo7 Alanaan's field effect is also only 8/12, it is easier for people using motocal when they add such skill with unique modifier via special effects rather than having to sum them up by had and use the other buff fields. Adding such effects can be helpful to see if a cap up weapon is better or not for example.

Element of additionalDamage is not necessary it gets calculated from the base damage instance, if it is dark damage or other elemental damage just changes the color of the number that pops up in game :)

h-yasha:
> Element of additionalDamage is not necessary it gets calculated from the base damage instance, if it is dark damage or other elemental damage just changes the color of the number that pops up in game :)

It does get effected by multiple stuff, like: elemental modf (like elemental superiority, check dark naru).

could look into "upgrading" additionalDamage in motocal in the future. but it's rather time consuming for little to mediocre.
could think of these TODO points for additionalDamage :
- element of Bonus DMG. would require looking into element of the character/"specific" elemental buffs
- separate instances for each additional damage (for calculating supplemental damage more accurately. since it's applied "fully" to each Bonus DMG instance/etc)

PR #259, opened 2019-06-14
Add/Update Various Support Abilities. Also fix Djeeta's DMG Cap EM

h-yasha:
**EDIT: note https://github.com/MotocalDevelopers/motocal/pull/259#issuecomment-509559488**

~~**Progress is reflecting my _local_ copy. I'll refrain from pushing any update until I get it "OK".**~~
- [x] [Yaia](https://gbf.wiki/Yaia): (Support 1) 5% boost to ATK ~~and 10% boost to DEF~~ for all other allies. 20% / 20% boost to critical hit rate for all other allies.
- [x] [Yaia (Holiday)](https://gbf.wiki/Yaia_(Holiday)): (Support 1) 20% / 20% boost to critical hit rate for all other allies when Yaia is a main ally.
- [ ] [Beatrix (Themed)](https://gbf.wiki/Beatrix_(Themed)): (Support 1) Boost to critical hit rate based on how low HP is (Only activates when below 60%). [reference](https://docs.google.com/spreadsheets/d/1tmd7bBBPkXXbIiecMDEZapdyhjvnshbNlAraOlzDSmY/edit#gid=298937121). Disabled: https://github.com/MotocalDevelopers/motocal/pull/259#issuecomment-509314539
- [x] [Lancelot and Vane](https://gbf.wiki/Lancelot_and_Vane): (Support 2) Increase Aegis-type weapon skills' 'Boost to max HP' effect by 30%. [reference](https://docs.google.com/spreadsheets/d/1tmd7bBBPkXXbIiecMDEZapdyhjvnshbNlAraOlzDSmY/edit#gid=1706595677)
  - Expected that it boosts all type of "normal" and magna HP boost weapon skill. Looking into:
    - [Titan's Aegis](https://gbf.wiki/Broadsword_of_Earth).
    - [Goldslime's Aegis](https://gbf.wiki/Goldslime_Bell).
- [x] [Sturm](https://gbf.wiki/Sturm): (Support 1) 80% bonus Fire damage for single attacks and 30% bonus Fire damage for double attacks.
- [x] [Levin Sisters](https://gbf.wiki/Levin_Sisters): (Support 2) 15% Bonus Light DMG effect upon triple attack.
[...]

PR #309, opened 2019-07-18
Fix composite support lookup logic for nest

kei-gbf:
- composite support can be nest now

- eachSupport function is moved to support.js

- Add unit test for flattenCompositeSupport
  - Limit maximum number of recursion (invalid case)

- Skip dupplicated support within a composite

Implementation details: (Flatten algorithm for nested data)

The flatten logic process the nested composite without recursive call.
The function does not consume function call stack,
make it possible by stack of generator object to process in a flat loop.

----

- [x] TODO: Fix unit test in TravisCI / Add polyfill for Generator


kei-gbf:
re-introduce the support module

- it was implemented for #259 and the patch is https://github.com/h-yasha/motocal/pull/8

that that module solve mainly two issues

----
### no duplicate clone code for isConsideredInAverage

```js
if (totals[key].isConsideredInAverage) {
    for (var key2 in totals) {
        totals[key2]["charaHaisui"] += charaHaisuiValue
    }
} else {
    totals[key]["charaHaisui"] += charaHaisuiValue
}
```

support logic has a lot similar code that if isConsideredInAverage for ...
then duplicated code for else case.

support_range.js can resolve it, processing isConsideredInAverage flag internally.

```js
for (let [name, chara] of range.all(totals, key)) {
    chara["charaHaisui"] += charaHaisuiValue;
}
```

----
### share support logic code by composite support definition

fall through approach in logic code will break when it's more than three cases.
composite support can solve the situation.

```js
// in global_logic.js
case "normalSupportKonshin_hpDebuff":
    totals[key]["HPdebuff"] += support.hpDebuff;
    // falls through
case "normalSupportKonshin":
```

it was quick/easy solution when implemented the "normalSupportKonshin_hpDebuff".
that can be

```
// in global_const.js
"normalSupportKonshin_hpDebuff": {
    "name": "...",
    "type": "composite",
    "value": [
        {ID: "normalSupportKonshin"},
        {type: "hpDebuff", range: "own", value: 0.15}
    ]
}
```

eachSupport() can generate two or more support for the composite type case.

the composite support sample is implemented only:
- mamorubeshi_sono_egao
- mamoritai_kono_egao

a lot support can be replaced with this module.
This patch provides the potential only, but I will not touch the support logic/const.
(testing a lot support change will take a time for review, I separate the task PR)

kei-gbf:
Problem: Travis report test failed

Details:
This problem CI failed was because of the Test environment gap.
test script run on the node engine, not same as production code they are run on browser.


Solutions:
- [ ] upgrade node version to 10.x
  - need to research issues, it require site admins to upgrade node too.
- [ ] Babel transpile or add polyfill plugin for Jest (Babel config)
  - I should choose this solution, it's only for test environment (devDependencies)

OrkunKocyigit:
I don't support node server upgrade over this change. Running on my node 8.11 server resulted in success even at Travis Failing build so I don't really know what is the issue. I would support babel solution as developer environments are easier to update/setup.

shinraminagi:
Hey, this commit broke the app: 1cc7a2c086f7ccc735b1204debb81d19abee87cf
You need 'require('./support')' in global_logic.js at least, but it cannot fix.

PR #312, opened 2019-07-22
Take LimitValues/DamageLimit out of damage calculation function 

### LINKED ISSUE #332: Graph for expected turn damage(player only) goes way beyond maximum possible damage cap on enmity
image
image

Images are self-explanatory, I believe.

kagekiyo7:
- Calculate LimitValues ​​outside of the function.
- Fix Enmty Graph/Damage Simulator damage
(I don't know a cause)

I can only write silly code, so I welcome PR or editing :)
image

purpose:
For Shiva summon buff.
For Chara's unique LimitValues of support abilitys.
For Displaying LimitValues.

## test
I have confirmed that PR is no different from master.
PR:
image
masetr:
image

## Fix Enmty Graph/Damage Simulator damage
As you can see above, the correct number is 1507489.
PR:
image
master:
image

closed #332


Groovman:
Best thing would be to put the limit values for ougis and such in the character data and patch the characters that have different limit values via the character script same as we do with DATA values and just grab the values from the character data. That way we wouldn't need the ougi cap support skills for Eternals and Lyria and such and could also patch other characters easier.

kagekiyo7:
True. But support skills are already registered. Unnecessary items will continue to remain in motocal.
Either way is fine for me.

kei-gbf:
just coding topic:

JSON parse/stringify is easy way but the function is called many times.

```js
JSON.parse(JSON.stringify(DEFAULTougiDamageLimitValues));
```

- __Problem:__ no need to re-copy the array many times.

```js
// another easy code for deep copy 2d array
// but I DON'T suggest this code, copying with map() is better.
const DEFAULTougiDamageLimitValues
DEFAULTougiDamageLimitValues.map(values => values.slice());
```

----
## Solution (suggest)

- [ ] define the variables out of function (should be in global_const.js)
  - [ ] const DEFAULTougiDamageLimitValues = ...
  - [ ] const DEFAULTnormalDamageLimitValues = ...
- [ ] clone with re-construct the limit values array

```js
// sample code
const xs = [[1,2], [3,4], [5,6]];  // assume this is default values. (declare this in global_const)
const ys = xs.map(([threshold,ratio]) => [threshold * (1.0 + ougiDamageLimit), ratio]); // new values

// XXX: "threshold","ratio" I am not sure this what name is suit.
```

for now idea&approach only, I can make the PR after GW event.

kagekiyo7:
Re test : OK

kagekiyo7:
## Existing issues

Enmity Graph: The numbers are close, but still different.
Damage Simulator: It was a misunderstanding that I thought it fixed.

 Normal Calculation
image
Enmity Graph
image
Damage Simulator
image

kagekiyo7:
Could someone review it?
This PR contains bug fixes.

kagekiyo7:
>Could you do the for loop in calcOugiDamage the same way you did in calcDamage?

Done.

PR #319, opened 2019-07-27
Display all party buff & Hide 0% party buff & Separate Field Effects

kagekiyo7:
## tasks 
- [ ] Display all party buff.
There were party buffs that is not displayed.
- [ ] Hide 0% party buff.
- [ ] Separate Field Effects.
Moved Field Effects(フィールド効果) to another line.
- [ ]  "フィールド効果” の切り分けに関して (about separate "field effects")
- 5か所に手を入れる必要がありますが、より保守しやすくする改善案があります。
now it requires 5 places to be modified. there is idea to improved it. (@kei-gbf will explain or PR later)


kei-gbf:
it should show NaN ? if not

```js
if (prof.normalBuff)
```

is enough, or

see also global_logic.js `module.exports.getTotalBuff`
they are checking by `!isNaN` for prof fields.

`NaN != 0` `undefined != 0` are true. zero check is useless here.

----

critical buff had no check zero cases.

- [x] no critical buff show nothing ... is ok
- [ ] ratio 0% case
- [ ] value 0% case

kagekiyo7:
I found a cause.
https://qiita.com/ma_me/items/9d2a72df209ae9832a3f

kagekiyo7:
OK. PR has been completed.

kei-gbf:
I misunderstood that still requires zero check then

```js
if (prof.xxx)
```

this code fill both zero or undefined/NaN etc checks.

- "", 0 -> false (zero check)
- null, NaN, undefined -> false (undefined etc checks)

----
```js
prof.criticalBuffCount > 0 && !isNaN(prof.criticalBuff)
```
no need !isNaN for this, because if it was NaN then NaN > 0 is false, no checked by isNaN.
isNaN check should be the first like, `!isNaN(value) && check value`

kagekiyo7:
Done.

kagekiyo7:
I test `if (prof.normalBuff)` :  
image

Perhaps this is a string.

kei-gbf:
- [ ] !isNaN check first? or second?

Generally, `A && B` and `B && A` are equals in logic, it's true.
however, there are two conclusions in this case.

- the order should should be `if the type was number(!isNaN) then check value (!= 0)`
- it's rare case to get NaN, so isNaN check later for optimization. 

both has merits/demerits.
so I would like to simple check

- if (parseFloat(value))
  - or parseInt

this code can treat "0" string case.

----
```js
var addPercent = (value) => intl.translate("percent", locale).replace("{}", value === undefined ? "0" : value);
```

I do not know what value assumed, this code check only "undefined" but null or NaN passed.
should be checked by !isNaN ?

kagekiyo7:
>so I would like to simple check

done

>should be checked by !isNaN ?

done

----
I tried to search, but isNaN does not check Null.
https://webllica.com/dont-use-isnan-javascript-numeric-check/

kei-gbf:
There was no cases to get null there,
when old save data then undefined,
or get initialized value. possible to get NaN by calc result.

kei-gbf:
- if (!isNaN(null) && null != 0) -> true ... Yes, we shouldn't do this
- if (null) -> false

when should avoid isNaN(null) is, the field was variant types
implicitly convert to zero as the URL article mention.
need to treat the null and 0 as different values.
in that case, we would put "null" as default value. can't check by isNaN.

src/result.js
```diff
@@ -489,7 +489,7 @@ var ResultList = CreateClass({
         if (parseFloat(prof.normalBuff)) buffInfo.push(intl.translate("通常バフ", locale) + addPercent(prof.normalBuff));
         if (parseFloat(prof.elementBuff)) buffInfo.push(intl.translate("属性バフ", locale) + addPercent(prof.elementBuff));
         if (parseFloat(prof.otherBuff)) buffInfo.push(intl.translate("その他バフ", locale) + addPercent(prof.otherBuff));
-        if (parseFloat(prof.otherBuff2)) buffInfo.push(intl.translate("その他バフ2", locale) + addPercent(prof.otherBuff2));
+        if (parseFloat(prof.otherBuff2)) buffInfo.push(intl.translate("その他バフ", locale) + "(2)" + addPercent(prof.otherBuff2));
```
kei-gbf: - many () for different things, some field use () for grouping values etc.
- make gap with profile.js use "その他バフ2"

kei-gbf: just add " " instead of "(2)"

commit log: "add separator between label and values"
make easy to understand why it's changed.

kagekiyo7: I could not group well.
`(buffInfo.push(intl.translate("その他バフ", locale) + "(2)") + addPercent(prof.otherBuff2));`
image

kagekiyo7: done.

kei-gbf:
addBlank does unnecessary tasks
- add % (by translate percent)
- remove % (by replace it again)

the function can be more simple:

```js
var addBlank = (value) => value || "0";
```

src/result.js
kagekiyo7: I was concerned about blank duplication(addPercent) in English, but it was fine.
image

kagekiyo7:
addBlank is intended to give blanks only in English.

image
image

kagekiyo7:
```js
    "percent": {
        "en": " {}%",
        "ja": "{}%",
        "zh": " {}%",
    },
```

kei-gbf:
the code intent mis-lead it add percent, remove percent.
unnecessary information.

it's time to all fields have the separator?, because

- ”別枠バフ10%, 別枠バフ2 10%" <- requires separator for label index and value
- ”別枠バフ: 10%, 別枠バフ2: 10%"

kei-gbf:
I have PR more three topics to this PR

- grouping critical buff
- reformat to "label: value"
- add function to push buff


(NOTICE: skip `#3` and label only once)
grop-critical

I still feel many braces but no ideas for now.

kagekiyo7:
I think that "Buff for Party"(title) and "XX Buff"(item) using the same separator(colon) doesn't look very good...
I can multiply and display OtherBuff1 and OtherBuff2.
What about doing so if only OtherBuff2 is a problem?
I think this style is the best.

`パーティ全体バフ: XXXバフ 0%, XXXバフ 0%,` or
`パーティ全体バフ: XXXバフ0%, XXXバフ0%,`
(If there is no blank between Buff and 0%, the break is perhaps easy to understand, only Japanese.)
`Buff for Party: XXX Buff 0%, XXX Buff 0%,`

kei-gbf:
The PR point is not about how displayed.
you can change the separator or change color etc, whatever.

- It's not good to have such exception cases, in a lot hard coded fields.
- Unify the format can reduce code size, simple code can be easy to maintenance.

kei-gbf:
Separator / Format are another topics
I got idea to do in component/CSS (idea approach only)

https://codepen.io/anon/pen/MNJwGd?editors=1111

- NOTE: `label: value,` format is widely use and well discussed in JSON, YAML formats.

----
Critical buff format, the format should be more improved:
- [ ] `(#1: 10%/10%, #2: 5%/5%, #3: 5%/5%)`
  - slash indicate "per" but it's also like a separator
- [ ] `(#1: 10%,10%/#2: 5%,5%/#3: 5%,5%)`
- [ ] `(#1 x1.1/10%, #2 x1.05/5%, #3 x1.05/5%)`
  - less confuse multiply and attack ratio

kagekiyo7:
Done. Thank you for the idea.

>it can be colored label like personal skills.

It might be nice, but I will go through for now.

src/result.js
```diff
@@ -479,22 +479,25 @@ var ResultList = CreateClass({
     
         var addPercent = (value) => value + "%";
         var addBlank = (value) => value;
+        // For avoid a "別枠バフ2XX%"
+        var addPercentAndSpace = (value) => (locale == "en" ? "" : " ")+ value + "%";
```
kei-gbf: it's being worse
- the maintenance cost of update code
  - should not have the "locale" dependency code in View (result.js)
  - doing the same twice in clone-code (code which copy-and-pasted)
    we should avoid such copy-and-paste code as possible, when update the code
    we have to search similar code and update them together.
    This is known code design problem, similar cause happened in #321 
    The cause is not just lack of tests, the code design issue.
- I've already mention the solution, unify the format can solve this code also become.
  whatever separator

kagekiyo7: done

kei-gbf:
- [ ] parseInfo / parseFloat について
  - DA/TA等で1%未満の値を扱う必要がないか (confirm: no need parseFloat for DA/TA ?)
- [ ] "フィールド効果” の切り分けに関して (about separate "field effects")
  - 5か所に手を入れる必要がありますが、より保守しやすくする改善案があります。
    now it requires 5 places to be modified. there is idea to improved it. (I will explain or PR later)

kagekiyo7:
>parseInfo / parseFloat について

小数に変換されるのはglobal_logic.js内なので大丈夫です。(profile内では整数)
```js
    if (!isNaN(prof.daBuff)) totalBuff["da"] += 0.01 * parseFloat(prof.daBuff);
    if (!isNaN(prof.taBuff)) totalBuff["ta"] += 0.01 * parseFloat(prof.taBuff);
```

kei-gbf:
念の為確認。該当箇所の UIで小数点以下が入る可能性、現状把握してる限りは、

- 直接入力可能になった場合 (fork等での運用を考慮)
- セーブデータを直接編集した場合

計算結果には小数点以下も反映、スキル量表示は正数のみとなります。
どちらも例外的な状況＆影響は軽微という事で、整数に統一。

Trade off:
- Float に統一した場合、下２桁を常に表示 => 表示幅問題。
- Int/Float を区別する場合。フラグ等を導入し対応は可能だが、追加で変更が必要。

----
global_logic.js での isNaN チェック
通常の利用で null が渡ることはないはずですが、潜在的には指摘されたような問題がありますね。
特に問題となっていなければ、不具合報告があるまでは現状のままで良いと思います。
セーブデータ編集で意図的に問題のあるデータを渡さない限りは大丈夫なはず。

kei-gbf:
I will re-review after Update branch

TODO myself
- [ ] re-list up review tasks (I almost forgot it had passed long time)

kagekiyo7:
## tasks 
- [ ] Display all party buff
There are party buffs that was not displayed.
- [ ] Hide 0% party buff
- [ ] Separate Field Effects
Moved Field Effects(フィールド効果) to another line.
- [ ]  "フィールド効果” の切り分けに関して (about separate "field effects")
- 5か所に手を入れる必要がありますが、より保守しやすくする改善案があります。
now it requires 5 places to be modified. there is idea to improved it. (@kei-gbf will explain or PR later)

kagekiyo7:
バフ数が多いと結構長くなってしまうので、○○バフのバフの部分削ってもいいかなと思ったのですが如何でしょうか？
image

----
何度もすみませんがやっぱり空白ないほうが区分けがハッキリしてて見やすいと思うんですよね…….。うーんレイアウト難しい。
image

PR #328, opened 2019-08-06
New Epic Weapons

Groovman:
### Weapons (Lv100 Stats on jp gbf-wiki)
* [x]   [Chains of Caucasus](https://gbf.wiki/Chains_of_Caucasus)
* [x]   [Savior of Hallowed Ground](https://gbf.wiki/Savior_of_Hallowed_Ground)
* [x]   [All-Might Battle-Axe](https://gbf.wiki/All-Might_Battle-Axe)
* [x]   [Unius](https://gbf.wiki/Unius)
* [x]   [Adamantine Gauntlet](https://gbf.wiki/Adamantine_Gauntlet)
* [x]   [Scales of Dominion](https://gbf.wiki/Scales_of_Dominion)

### Weapon Skills to implement:
* [x] Grand Epic (Chains of Caucasus, Unius, Adamantine Gauntlet)
* [x] Staff Resonance (Savior of Hallowed Ground)
* [x] ~~Amber Arts (All-Might Battle-Axe)~~
* [x] Heroic Tale (All-Might Battle-Axe)
* [x] ~~Graphite Arts (Scales of Dominion)~~
* [x] Absolute Equality (Scales of Dominion)

PR #329, opened 2019-08-07
Clean up clone code by has_3rd_skill flag

kei-gbf:
This change will reduce the coding maintenance cost.
Nothing change/effects to armData.json

### Problem:

has_3rd_skill flag made large clone code.

When we want to add new category(series key) field, on #328 
we need to modify the same twice on the duplicated codes.

### Solution:

I've removed the clone code and resolve the index gap.

we can read the next value directly `next(iterator)`
that does not consume index counter by `enumerate(iterator)`
at the result, it will keep index number => no need clone code.

----
## Checklist

- [x] the generated armData.json match perfectly with current armData.json

```
$ wc armData.json
 23405   47263 654154

$ wc armData.json.bak
23405   47263 630749
```

the size diff are my local repo setting, git will convert autocrlf on commit
so the actual size will be 654154(bytes) - 23405(lines) = 654154(bytes) matched

check it by diff --strip-trailing-cr ... ok

-----
### Other topics

I found issue but this PR will not fix (report only)

- [ ] element check
  - Python string.find return zero based index, -1 for not found
    - e.g.  "ABC".find("A") => 0   "ABC".find("D") => -1
   - Omega weapons:
       - txt_source has `|火|'
       - `if value.find("火") > 0:` fail because find() return 0 
       - then Omega weapons txt_source has "火"(fire) but it's "dark" in armData.json
          - is it a bug or expected behavior? anyway the use of find() should be ">= 0"

I won't fix it in this PR, because this will change armData.json
it's mix topics make hard data checks.
I want to ensure this PR changes nothing to JSON data.

- [ ] Clean up topic: no need the long else block,
       It's not good style to have such deep indent in python.

```python
if row_length <= 1:
    continue
else:
    ...
```

should be

```python
if row_length <= 1:
    continue
...
```

I won't fix it in this PR, because the indent-only change covers and hides the diff
that make hard to find mainly stuff I changed in this PR.

### DIFF
--- a/scripts/arm_data_converter.py
+++ b/scripts/arm_data_converter.py
@@ -1042,19 +1042,20 @@ def processCSVdata(csv_file_name, json_data, image_wiki_url_list, image_game_url
 
     for row in mycsv:
         newdict = OrderedDict()
-
         row_length = len(row)
         if row_length <= 1:
             continue
         else:
             # Row Lengths Overview
+            #  8: 2nd + 3rd skills
             # 17: 3*
             # 19: 4*
-            # 20: 4* + 3rd Skill
+            # 20: 4*
             # 21: 5*
-            # 22: 5* + 3rd Skill
+            # 22: 5*
             has_3rd_skill: bool = row_length == 20 or row_length == 22
-            for index, value in enumerate(row):
+            row_iter = iter(row)
+            for index, value in enumerate(row_iter):
                 if index == 1:
                     m = key_pattern.search(value)
                     if m:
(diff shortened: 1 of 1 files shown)


Groovman:
>   * Python string.find return zero based index, -1 for not found
>     
>     * e.g.  "ABC".find("A") => 0   "ABC".find("D") => -1
>   * Omega weapons:
>     
>     * txt_source has `|火|'
>     * `if value.find("火") > 0:` fail because find() return 0
>     * then Omega weapons txt_source has "火"(fire) but it's "dark" in armData.json
>       
>       * is it a bug or expected behavior? anyway the use of find() should be ">= 0"
> 
> I do not fix this in this PR, because this will change armData.json
> it's mix topics make hard data checks.
> I want to ensure this PR changes nothing to JSON data.

Does this affect everything else than Ultima/Atma Weapons? Ultima/Atma Weapons have a choice for their element anyway. We should maybe change them to All in the txt so they show up in all categories anyway instead in just Dark.

kei-gbf:
> Does this affect everything else than Ultima/Atma Weapons? 

Utima/Atma weapons only, I will show the list on the bottom of this comment.

I found three topics on the element check code.

- [ ] Ultima/Atma Weapons has "fire" as default, __should be "all"__
- [ ] str.find() > 0 checks __should be >= 0__
  - there are no other cases except Ultima/Atma weapons,
    so after fix their element to "all", no actual problems with `> 0`
    but it's better correct them to show the code intent.
- [ ] __the element check should have "闇" for "dark" explicitly.__
  currently else-block set "dark" element as default.
  the else-block has both "dark" case and error cases.

```python
else:
    newdict["element"] = "dark"
```

----
report non-dark element but set dark as default

(this is a temporary customize for reporting)
```python
elif value.find("闇") > 0:
    # catch the correct "dark" case here, not ">=" yet to pass other 0 cases.
    newdict["element"] = "light"
elif value.find("全属性") > 0:
    newdict["element"] = "all"
else:
    # for logging, "火".find("火") > 0 is false and reach here
    print(" - ", value, newdict["ja"])
    # temporary set "none", because skill use default element.
    newdict["element"] = "none"
```

script reported:

 -  火 オメガスウォード
 -  火 オメガバゼラード
 -  火 オメガバルーチャ
 -  火 オメガファルシャ
 -  火 オメガロッド
 -  火 オメガシューター
 -  火 オメガクロー
 -  火 オメガボウ
 -  火 オメガストリング
 -  火 オメガブレイド
 -  火 無垢なる竜の剣
 -  火 無垢なる竜の短剣
 -  火 無垢なる竜の槍
 -  火 無垢なる竜の斧
 -  火 無垢なる竜の杖
 -  火 無垢なる竜の筒
 -  火 無垢なる竜の爪
 -  火 無垢なる竜の弓
 -  火 無垢なる竜の弦
 -  火 無垢なる竜の刃

Groovman:
> Ultima/Atma Weapons has "fire" as default, should be "all"

Replace "`|火|" with "|&color(#ff9900){全属性};|" in armData-ssr.txt, Done. Did it in a Commit in PR #328.

kei-gbf:
> Replace "`|火|" with "|&color(#ff9900){全属性};|" in armData-ssr.txt, Done. Did it in a Commit in PR #328

Can we do this separate PR ? 
I can fix other python script issues.

This PR change is required for #328
I just separated for the report does not mess Epic Weapon topic.

Others are not epic weapon related issues.
they can be merged independently.

and other side, epic topic does not need to have those changes.
the diff will contains those changes, it unnecessary for the 328.

so I will move the commit

- [ ] new PR for python scripts
- [ ] move the commit (cherry-pick and revert)

PR #335, opened 2019-08-13
Stackable Support Abilities

h-yasha:
This PR will include Support Abilities that has varied number of values based on stacks like Hallessena's Downright Crazy! or other "simple" values like Athena's Searing Crimson.
_target is Burned condition and likes (based on target condition) could be generalized later on (and added beside Sun-Touched Paradise/etc._
 
- [x] Add Additional field for "stage" value select in chara.js.
- [ ] implement logic. (not so "good" implementation.)

### Fire
- [ ] [Aliza](https://gbf.wiki/Aliza): Everstrong Spirit
- [ ] [Athena](https://gbf.wiki/Athena): Searing Crimson
- [ ] [Fraux](https://gbf.wiki/Fraux): Lonely Figure
- [ ] [Ghandagoza](https://gbf.wiki/Ghandagoza): Unrivaled Fighter
- [ ] [Grea](https://gbf.wiki/Grea): Flame Arts
    - [ ] also include C.A. innate effect.
    - [ ] _Uptime_ select. (besides on/off)
- [ ] [Heles](https://gbf.wiki/Heles): Scarlet Ribboned Lance
- [ ] [Ilsa (Summer)](https://gbf.wiki/Ilsa_(Summer)): Rise Again
- [ ] [Ilsa (Summer)](https://gbf.wiki/Ilsa_(Summer)): Summer Hell Drill Sergeant
- [ ] [Io (Summer)](https://gbf.wiki/Io_(Summer)): Heat Haze
- [ ] [Kallen Kouzuki](https://gbf.wiki/Kallen_Kouzuki): Zero Squad Commander
    - Not a priority (C.A. reset stacks) 
- [ ] [Magisa](https://gbf.wiki/Magisa): Almighty Sorceress
- [ ] [Magisa](https://gbf.wiki/Magisa): Morax, the Demon
- [ ] [Metera](https://gbf.wiki/Metera_(Fire)): Genius
- [ ] [Metera](https://gbf.wiki/Metera_(Fire)): Complete Arcane Bow
- [ ] [Rackam](https://gbf.wiki/Rackam_(Grand)): Sleight of Hand
- [ ] [Rackam](https://gbf.wiki/Rackam_(Grand)): Punish Save
- [ ] [https://gbf.wiki/Shiva](https://gbf.wiki/Shiva): Dance of Nataraja
    - will replace currently existent support ability
- [ ] [Tien](https://gbf.wiki/Tien): Treacherous Trigger (Lv. 90)
[...]

PR #342, opened 2019-08-16
Add Turn Counter

h-yasha:
- [x] Add UI element.
- [x] Applied Turn effect to (Normal/Magna) Element ATK L.
    - currently other than sklv 20 values are unknown? currently it's defaulted to have sklv 20 value. however in this PR it has been changed to be 0 other than sklv 20.
- also to be used in #335 as various support abilities have turn count interaction.

- [ ] interaction with Damage Simulator.


h-yasha:
- [ ] merge with #352

kagekiyo7:
I updated japanese translation.

kagekiyo7:
> interaction with Damage Simulator.

This is probably an advanced requirement, so I'd better do in a different PR.
(To remember, you should write in Issues.)

PR #353, opened 2019-09-26
Display min weapon in armtab & Display "本" on arm template

kagekiyo7:
image

image


src/armlist.js
```diff
@@ -231,6 +231,7 @@ var ArmList = CreateClass({
                                     <Panel.Title toggle>
                                         {(ind + 1)}: {(alist[ind] != null) ? alist[ind].name : ""}
                                         &nbsp; {(alist[ind] != null && alist[ind].name != "") ? alist[ind].considerNumberMax + "本" : ""}
+                                        {(alist[ind] != null && alist[ind].name != "" && alist[ind].considerNumberMin) ? "(" + intl.translate("最小", locale) + alist[ind].considerNumberMin + "本" + ")" : "" }
```
OrkunKocyigit: Can you move "本" to translation key? It can be empty string as English does not require appender to point to origin.

kagekiyo7: done.

PR #358, opened 2019-09-28
Add new sort calc "Total DMG in Simulate 100T ÷ 100"

kagekiyo7:
Caluclation 100T damage  based on Granblue specification.
It can more accurater caluclation in Kengo party etc.
And, also calculate turn effect buff in future customization.
I think that motocal should change to this calculation method that is closer to the specifications of granblue rather than the current calculation method.

Problem:
- Processing is little slower.
- Don't know calculation is correct. (can't confirm)

It can increase number of turn if it can do faster processing. 
Someday TODO: Add support for multi sled. (impossible?)


kagekiyo7:
I can't come up with the correct implementation of chainburst.
The current is wrong.

kagekiyo7:
typo, correct:
Change djeeta's default ougiGage to 100%

kagekiyo7:
Memo:
Gaining negative ougi gage decreases number..
https://twitter.com/zekasyuz/status/846696279417643010

kagekiyo7:
DATA random number test(result is not stable.):
https://github.com/kagekiyo7/motocal/blob/Add-new-calc-DATA-random/src/newcalc.js#L40

kagekiyo7:
Add lockoutTime test:
https://github.com/kagekiyo7/motocal/commit/0b4b31dbe8dee89ffd29b4ddcb979238575e931e

kagekiyo7:
I’m planning to rebase.

PR #361, opened 2019-09-29
Add filtering feature for chara template

kei-gbf:
This patch add 4 more fields for chara filtering.

## Checklist
- [x] filter function
  - [x] chara sex
  - [x] chara race
  - [x] chara fav (favorite weapons)
  - [x] chara type

not tested so much, especially multiple value like "human/erune"

chara-filter

## Won't do in this PR

- Weapon filtering by type or series ... should be in #331 or other.
- layout: need label for fields ?
- layout in mobile
- users may want "Reset" button
- Known Issue: armType "none"
- multi select

----
## Internal changes

- global_const.js
  - make utility functions to generate option list for each locales
    オプションリストを生成するユーティリティ関数を定義
    - _generateFilterOptions
    - _setupFilterOptions
  - fix naming case filterelements -> filterElements
- template.js
  - share common elements which used in mobile and PC
    モバイル用とPC用で表示ロジックが分離しているので、
    共通部分をまとめる。
  - separate filter and map functions
    絞り込み機能と表示の為の要素生成機能を別関数に整理して分離
  - refactoring event handler, setState
    with ES6 Enhanced Object Literals


src/template.js
```diff
@@ -139,99 +143,89 @@ var RegisteredChara = CreateClass({
...
         });
     },
     clickedTemplate: function (e) {
         this.props.onClick(this.state.charaData[e.target.getAttribute("id")]);
     },
     handleEvent: function (key, e) {
-        var newState = this.state;
-        newState[key] = e.target.value;
-        this.setState(newState)
+        this.setState({[key]: e.target.value});
```
kei-gbf: NOTE: update state better way with ES6 Enhanced Object Literals
react will merge, no need full state object.

kei-gbf:
- [ ] layout broken, footer was not displayed within border

kagekiyo7:
> layout in mobile

I did it instead.

original:


updated:

kei-gbf:
- [ ] TODO: 性別不明?の欄が空欄だった

OrkunKocyigit:
I only tried tiny bit.
- [ ] [最終]サラーサ (Threo) defined female yet appears on male filter.
image

kei-gbf:
Thx for report
it's a bug seems "male" had matched fe(male)

```js
    return ([key, val]) => (
        (filterElement == "all" || val.element == filterElement) &&
        (filterText == "" || val[locale].toLowerCase().indexOf(filterText.toLowerCase()) != -1) &&
        (filterRace === "all" || val.race.includes(filterRace)) &&
        (filterType === "all" || val.type === filterType) &&
        (filterFav === "all" || [val.fav1, val.fav2].includes(filterFav)) &&
        (filterSex === "all" || val.sex.includes(filterSex)));
```

- [ ] filterSex: male (user selected)   val.sex (charaData)
  - "include" assumed "male/female".include("male") work
  - [ ] but not for "female".include("male") this is a bug.
  - [ ] and it was missing reverse case "male".include("male/female")
- [ ] filterRace also has similar code

----
Research for the solutions:

- using Set data type ... if JavaScript support the data type
- split by "/" and Array.include
  - but it will make temporary array that unnecessary most cases
- regex match
  - "male/female".match(/\bmale\b/) true
  - "female".match(/\bmale\b/) will be false

OrkunKocyigit:
It is impossible to select any characters in this branch.

src/template.js
```diff
@@ -139,99 +170,85 @@ var RegisteredChara = CreateClass({
...
+            <FormControl componentClass="select" value={this.state.filterFav}
+                         onChange={this.handleEvent.bind(this, "filterFav")}>{selector[locale].filterFavs}</FormControl>
+            </>;
+
+        const filterFunc = _generateCharaFilterFunc(this.state, locale);
+
+        const mapFunc = ([key,val]) =>
+            <div className="onechara" key={key}>
+                <p>{val[locale]}</p><br/>
+                <Image rounded onClick={clickedTemplate} id={key}
```
OrkunKocyigit: clickedTemplate will be always undefined. It needs to be set during render or referenced in initialState or replaced with this.clickedTemplate

kei-gbf: thank u for report, seem I did not do the acceptance test.
I see that context of the closure. is it because of JSX component context?

- key, val ... are arguments ok
- clickedTemplate ... outer scope variable (different context in run-time)
- locale ... also undefined ?

I will try this.clickedTemplate as quick solution, and also fix "locale" if needed.
I think ideally those closure should be designed as testable function. (future issue)

kei-gbf: my misunderstood, it was not context issue.
at the head of function, I was trying to import clickTemplate local variable from "this.state"
locale was ok, const clickedTemplate was undefined

kei-gbf:
Official Game updated filter/sort
I would not plan the feature at this PR.
weapon filter also use same UI, the multi-select feature should be another PR if needed.

## This PR will not include
- [ ] UI allow multiple select
- [ ] filter logic to support multiple token
  - currently internal logic support matches single with multi token like "sword/dagger"
  but not the case both token are multi.

PR #293, opened 2019-07-08
Add "Speed up charge bar during one-foe attacks" of masterBonus

kagekiyo7:
"Boost to main weapon's ATK when main weapon is ***" is impossible because motocal does not have a main weapon concept.

Add support for "Speed up charge bar during one-foe attacks". (通常攻撃時奥義ゲージ上昇)

PR #365, opened 2019-10-13
Add support for Math.ceil of  expected ougi gage

kagekiyo7:
test:
TA 100%, Charge Bar Buff 50% 
→ 37.0x1.5=55.5% → expected ougi gage(Charge Bar Rise per Turn) 56%
image


OrkunKocyigit:
I have issue understanding this formula as a general.
If I have character with %100 TA and 0% buff then in game I would gain 30% charge bar every turn right? Where does the 37% coming from?

kagekiyo7:
> I have issue understanding this formula as a general.

It's bonus.

> Double Attacks give a bonus of 2% and Triple Attacks give a bonus of 7%. A Single Attack therefore generates 10% charge bar, a Double Attack 22% and a Triple Attack 37%. This bonus is affected by Charge Bar speed up buffs.
> https://gbf.wiki/Charge_Attack

OrkunKocyigit:
LOL I never realized past 2 years I have been playing. I'll try check this PR later.

OrkunKocyigit:
I tried to simply formula via wolfram but it says it is as good as gets. Tests look fine. Thank you for your work

kagekiyo7:
Thank you for the review. I would like this when there is time.
#293

PR #375, opened 2019-11-11
Zwei Schaedel 2nd Skill

OrkunKocyigit:
**Scope of Change:**

This PR aims to implement Zwei Schaedel second skill provides %20 echo to 3rd hit.

**Requirements:**

- Bonus dmg effect **does not** stack with multiple weapons.
- Bonus dmg effect only applies to **3rd** hit of triple attack.
- Bonus dmg effect only applies to **dark** allies.

**Changes included in this PR**
- [ ] Weapon parser python script now parses Zwei Schaedel 2nd skill.
- [ ] 2nd skill removed from armData-ssr.txt for 3* weapons.
- [ ] Zwei Schaedel weapon now has "Bonus Damage on 3rd Attack" (3回目の攻撃に闇属性追撃効果 (20%)) skill.
- [ ] zwei-echo sub skill has been added.
- [ ] New skill type echoThirdHit added to cover bonus damage on 3rd attack.
- [ ] zwei-echo adds 0.20 to echoThirdHit effect added to skill calculations.
- [ ] echoThirdHit effect has been added to additional damage with given value amount / 3.

**Changes NOT included in this PR**
- Any other weapon

**Use Cases**
- Dark char %100 TA will have 20 / 3 = 6.7% bonus echo at average.

image

- Non dark char will have 0% echo regardless.

image

- Dark char with 10% TA should have 20 * 0.10 / 3 = 0.67% bonus echo at average.

image

- Any char with 0% TA should have 0% bonus echo.

image


kei-gbf:
note related tasks (noted on #233 as uncompleted task)
> Need to account for Bonus DMG in supplemental damage as it applies to it separately. (Add an echo count field for Djeeta and each character?)

- [ ] supplemental damage for additional(bonus) damage implementation was not completed
  - multiple bonus damages should get multiple supplemental damages,
     but only single supplemental damage for multiple bonus damage.
 
----

追撃枠が未考慮の為、与ダメ上昇の計算に影響有
複数の追撃に対して与ダメ上昇が一つしか乗りません。
(#233 与ダメ上昇実装時からの課題）

PR #377, opened 2019-11-12
[2019-11-08] New Moon Weapons

OrkunKocyigit:
**Scope of Change:**

This PR aims to add following weapons.
Taming Sari: https://gbf.wiki/Taming_Sari
Nereus Staff: https://gbf.wiki/Nereus_Staff

**Requirements:**

- Taming Sari should be added with attributes described in wiki
- Nereus Staff should be added with attributes described in wiki
- Proteus's Prophecy effect needs to be implemented

**Changes included in this PR**
- [ ] armData updated to include 3 and 4* weapons.
- [ ] Parser now parses "Tsunami's Progression".
- [ ] Parser now parses "Proteus's Prophecy".
- [ ] Proteus's Prophecy now implemented as NormalHP += Num of equipped staff * 2

**Changes NOT included in this PR**
- Tsunami's Progression is same skill as opus weapon key so it is defined as opus skill. It might be refactored into more general skill.


src/global_logic.js
```diff
@@ -1853,6 +1853,14 @@ module.exports.addSkilldataToTotals = function (totals, comb, arml, buff) {
...
                         // skill is all allies not restricted to element
                         totals[key]["criticalDamageLimit"] += comb[i] * 0.05;
+                    } else if (stype == 'wand-count') {
+                        let numOfWand = 0;
+                        for (let i = 0; i < arml.length; i++) {
+                            if (comb[i] > 0) {
+                                numOfWand += arml[i].armType === "wand" ? comb[i] : 0;
+                            }
+                        }
+                        totals[key]["normalHP"] += amount * numOfWand;
```
kei-gbf: - [ ] Naming confuse point, epic weapon car-ong wand used stype: "epic" amount: "count-wand"
  - and outer scope has variable "countWand" it's unused if no car-ong wand set,
    we can set the variable before for-loop, then both skill can share the wand count info.
- [ ] reuse epic.countWandType function
  - ` totals[key]["normalHP"] += amount * epic.countWandType(arml, comb);` (not tested)
  epic module has generic function to generate another weapon types count.
  it assume future skill implemented count sword types etc.

OrkunKocyigit: Changed on 42da935

armData.json
```diff
@@ -19058,6 +19058,86 @@
...
+        "element": "light",
+        "type": "dagger",
+        "skill1": "normalL",
+        "skill2": "non",
+        "element2": "light",
+        "skill3": "non",
+        "element3": "light",
+        "minhp": 38,
+        "minattack": 410,
+        "hp": 259,
```
kagekiyo7: >"hp": 259,

235 is correct.
imege from internet.
image

PR #393, opened 2019-12-13
Move calculating results off of main thread

Groovman:
See: https://github.com/Groovman/motocal/pull/9


kei-gbf:
- [x] window.navigator.userAgent.toLowerCase() -> self.windows
  changed to "self" for worker ... ok
- [ ] now _ua can be null ? 
  - access to _ua object may be Error
  - not happen in browser, but test is in nodejs

one of reason the _ua's anonymous function took userAgent string as argument  is
look like remove the dependency to window object, that non-browser environment
may not have. ideally it should be out of the module, caller's environment should have it.
user-agent is a dynamic information, not const in fact.

Groovman:
If the internet exploder can do it you need no polyfill :)

kei-gbf:
it's should be ok, the most of modern browser support web-worker now and
the Motocal top-page said support target is Chrome.

There is an optional idea that if anyone wants previous implementation then
build option to enable the worker (webpack preprocess)
this provide a merit if the worker code had problem in some environments
site admin can rebuild to disable worker code, without back to find a stable repository.

----

_ua ... the behavior changes
using "self" seem changes for worker
but calculate_result_worker.js did not use "_ua"

kei-gbf:
BTW I saw the fix translation commit but different hash
we need that standalone PR ? instead of every PRs has the patch.

- [ ] new PR for fix translation
- [ ] revert fix translation from PRs and merge from master
  Because of the PR's diff contains unnecessary patch.

Groovman:
> BTW I saw the fix translation commit but different hash
> we need that standalone PR ? instead of every PRs has the patch.
> 
> * [ ]  new PR for fix translation
> * [ ]  revert fix translation from PRs and merge from master
>   Because of the PR's diff contains unnecessary patch.

No I am tired of fixing that translation issue, every time I have to deploy and it gets back merged in. I am not doing an extra PR for 2 wrong letters, when we also squash PRs anyway so that patch is not even listed in the commit log in the end. Also I want to have a merge conflict when it gets merged in again.

kei-gbf:
not listed in commit log ... ok
but the diff is stay on PR's diff, other PRs also has the same.
I was going to suggest avoid those same fixes on every PRs

Groovman:
If that should go into prod, the abuse of var instead of let in most functions has to go, at least in game_logic.

kei-gbf:
計算処理をWebWorkerで処理します
利点としては、多くの武器を計算しようとした時にブラウザが固まるのを回避できる（要ストレステスト）
課題は、もし対応外のブラウザのサポート要求が多かった場合、戻すのが困難。

This PR implement worker for calculation
The merit is when users want to calc many weapons, it expect to avoid browser freeze
(we need stress test)
issue is if a lot users require to support old browser non-worker version, it's hard to roll back.

----
I have several ideas for refactoring calculate_result_worker.js
but the code was simple move from result.js
another PR can do that task, because of the diff
(move and modify once is not easy to compare diff)

kei-gbf:
Anyone can stress test this PR ?

worker resolves browser freeze when calc too many combination :+1:
but the performance may go worse when many workers are launched.
can we limit the number of worker

src/global_const.js
```diff
@@ -78,7 +82,7 @@ module.exports._ua = (function (u) {
             || u.indexOf("ipod") != -1 || (u.indexOf("android") != -1 && u.indexOf("mobile") != -1)
             || (u.indexOf("firefox") != -1 && u.indexOf("mobile") != -1) || u.indexOf("blackberry") != -1
     }
-})(window.navigator.userAgent.toLowerCase());
+})();
```
kei-gbf: `navigator` is accessible from either main or worker thread
https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope

I suggest
```js
function (u){
    return {/* snip */]
})(navigator.userAgent.toLowerCase() || "");
```

reason:
- remove browser dependency from function body (but not from module)
- an empty value as undefined-case argument (non-browser environment)
  - reduce maintenance cost `self.window` twice

NOTE: pls no touch Table detect code, can't device tests in this PR (not refactoring time)

src/result.js
```diff
@@ -40,141 +40,32 @@ var {
...
-            return {summon: summon, result: res, sortkeyname: sortkeyname, totalItr: totalItr}
-        } else {
-            return {summon: summon, result: []}
-        }
+    calculateResult: function ({ profile, armlist, summon, chara, sortKey }) {
+      const { previousArmlist, previousCombinations } = this.state;
+      this.setState({
+        calculatingResult: true
+      });
+      const worker = new ResultWorker();
```
kei-gbf: Current code has memory leak in many weapon and change often.
and worker does not promise the order of receive results.

1. store `worker` instance to global (property of `document` or `window`)
  (no need to care lock or atomic here, it's thread-safe action in main thread)
2. kill previous worker by invoke terminate method
  https://developer.mozilla.org/en-US/docs/Web/API/Worker/terminate
3. spawn new worker to global place. (see step 1)

This is not enough efficient solution, terminate() kills worker, we can't reuse the previous worker resource.
it's a limitation by the design of the worker. at least I expect it resolves memory leak.

kei-gbf:
@Groovman 

it's now broken, the change should be after merged ?
it make this PR untestable.

err

- revert 726f2a0
- after merged
- revert the revert (git can rollback by revert the revert commit)

Groovman:
Please merge, it is tested for months now.

src/result.js
```diff
@@ -1156,23 +1149,27 @@ var Result = CreateClass({
...
+                                        <span
+                                            className={"label label-default"}>{intl.translate("実質奥義上限", locale)}</span>&nbsp;
+                                        {formatCommaSeparatedNumber(ougiDamageRealLimit)}&nbsp;
                                     </span>
-                                </div>
+                            </div>
                         );
-                    };
+                    }
+                    ;
```
kei-gbf: I had not seen this, what formatter has this rule?

src/result.js
```diff
@@ -33,38 +33,30 @@ var enemyDefenseType = GlobalConst.enemyDefenseType;
...
+        }
+        worker.onmessage = ({data: {result, previousArmlist, previousCombinations}}) => {
+            this.setState({
+                result,
+                previousArmlist: previousArmlist || this.state.previousArmlist,
+                previousCombinations: previousCombinations || this.state.previousCombinations,
+                calculatingResult: false
+            });
+            worker.terminate();
+            worker = undefined;
```
kei-gbf: terminate() onmessage will not solve memory leak issue?

Groovman: It will, the memory leak happens because workers have message listeners on them. So the garbage collector can't destroy them and free memory. Terminate() closes the listeners from the outside main thread. The worker can be stopped after the result got returned on that callback.

kei-gbf: there were two topics,
It's ok about solve the memory leak when task was finished. ( tested :ok: )
How about memory usage on when many worker running
the situation like #404 reported still can be happen.

no memory leak but kind of resource exhaustion.

Can I approve this PR first? then I will report that as another issue.
at least works good on general use.

Groovman: Memory is less of an issue than the computation time. If you have many many combinations, you will run in bottleneck issues regardless of that new solution or the old code. The bottleneck is just not as worse as we currently have with everything running on the main thread and all the computation locking the user.

The best solution would be using a thread pool from like threads.js, but we still have to what is already done here making the calculation into a worker function. But for the current state the worker solution is enough, they should not run that long.

PR #394, opened 2019-12-14
Chart fixes and en wiki nvd3 chart export

Groovman:
Some fixes for charts translation and chart export for en wiki nvd3chart format export.

----
- Add support for export nvd3 chart export.
image

- Give 2nd skill name of Beast Weapons(四象武器) when naming.
e.g. Baihu Claw **Pontus**.


src/translate.js
```diff
@@ -1754,7 +1759,7 @@ var multiLangData = {
         "zh": "加入图表",
     },
     "本": {
-        "en": "",
+        "en": "x",
```
kagekiyo7: There are three places use key of "本". Is there no problem? (just confirm)

image
image
image

src/chart.js
```diff
@@ -169,9 +235,15 @@ var HPChart = CreateClass({
                             {selector[locale].supported_chartsortkeys}
                         </FormControl>
                         <span>CSV</span>
-                        <CSVLink className={"btn btn-default"} data={csvData}><span
+                        <CSVLink className={"btn btn-default"} data={csvData} filename={"motocal_charts.csv"}><span
+                            className={"glyphicon glyphicon-download-alt"} aria-hidden={"true"}></span></CSVLink>
+                        <button className={"btn btn-default"}><CopyToClipboard text={csvClipBoardData}><span
+                            className={"glyphicon glyphicon-copy"} aria-hidden={"true"}></span></CopyToClipboard>
+                        </button>
+                        <span>WIKI</span>
```
kagekiyo7: "WIKI" should change to "NVD3" to make it more specific. (japanese wiki can't use this)

src/translate.js
Groovman: Yeah set names in chart section are looking strange with just a number at the end and 1x would be one time in western same as  1本.

src/chart.js
Groovman: Done

kagekiyo7:
I added a description without permission on top.

Groovman:
Technically already got used https://gbf.wiki/User:Midokuni/Notepad/MotocalExport/Colo :)

PR #395, opened 2019-12-16
Caim 2nd support and some fixes for epic weapons calculation

Groovman:
Hey, 

epic weapons calculation had an issue when changing epic weapons elements, I dumbed down the calculation but is now more stable. I also implemented Caim's 2nd support ability and moved the support ability calculation on to the combination calculation part.

If you want to add other support abilities that require armlist or combination after this PR they are available in the support calculation.


scripts/arm_data_converter.py
```diff
@@ -1008,16 +1008,16 @@
 
 # New epic weapons
 skillnamelist["epic-grandEpic"] = {
-    u"エピックブランド・ゲイン": "all",
+    u"エピックブランド・ゲイン": "dark",
```
kei-gbf: is this "dark" ? 2nd skill does not have element

src/global_logic.js
```diff
@@ -2816,6 +2800,24 @@ module.exports.treatSupportAbility = function (totals, chara, buff) {
...
+                case "hanged_man_reversed":
+                    if (index > 4) {
+                        if (epic.isAllUniqueArm(arml, comb)) {
+                            if (totals[key].isConsideredInAverage) {
+                                for (var key2 in totals) {
+                                    totals[key]["caimOther"] += support.value;
+                                    totals[key]["normalDamageLimit"] += support.value;
+                                    totals[key]["ougiDamageLimit"] += support.value;
+                                }
+                            } else {
```
kei-gbf: support_filter module now provides iteration support_filter.range.element["earth"]
no more duplicated code for non-isConsideredAverage
(the iteration function return self if isConsideredInAverage was off)

```js
for (let [name, chara] of range.element["earth"](totals, key)) {
    chara["caimOther"] += support.value;
    chara["normalDamageLimit"] += support.value;
    chara["ougiDamageLimit"] += support.value;
}
```

EDIT: "chara" was used in outer scope, "let" protect with in the scope though,
another variable name for avoid the confuse.

scripts/arm_data_converter.py
Groovman: because all is reserved for weapons whose element gets changed when added to the arm list. The element selector gets into a wierd state when you throw "all" at him, it shows fire but fire is not selected. For non elemental stuff the default element is dark, so the selector is set to dark and not in an error state.

src/global_logic.js
Groovman: It is set up how other skills are set up. I am not doing let and const stuff which behaves differently from other stuff.

kei-gbf: they are an old code. the improve points are
- deep unnecessary indent
- duplicated code for non-isConsideredInAverage  (it make double maintenance cost)

```js
if (index > 4 && epic.isAllUniqueArm(arml, comb)) {
    for (let [name, chara] of range.element["earth"](totals, key)) {
        chara["caimOther"] += support.value;
        chara["normalDamageLimit"] += support.value;
        chara["ougiDamageLimit"] += support.value;
    }
}
```

Groovman:
Moved to #399 to prevent merge issues.

PR #399, opened 2019-12-21
Recent patches till 2019 12 19

Groovman:
All new characters, weapons, changes and fixes till current patch.

Levi Malice weapon data missing for lvl 100, but it is just chaning 4 values in the csv if they are getting updated.

My last PR.

PS. merged in the caim and epic fixes to ensure github can't fuckup merging both.


src/global_logic.js
```diff
@@ -1692,8 +1694,31 @@ module.exports.addSkilldataToTotals = function (totals, comb, arml, buff) {
                         } else if (skillname == 'cosmosPC' && typeCharaContains("pecu", totals[key])) {
                             totals[key]["cosmosDebuffResistance"] = comb[i] * 20.0;
                         }
-                    } else if (stype == 'cosmosArm') {
-                        // Skip Cosmos Weapons Skill
+                    } else if (stype == 'cosmosLimit') {
+                        // Cosmos Weapons Damage Limit Skill
+                        if (amount == 'sword') {
+                            totals[key]["normalDamageLimit"] += comb[i] * 0.01 * Math.min(10, epic.countSwordType(arml, comb));
+                            totals[key]["ougiDamageLimit"] += comb[i] * 0.01 * Math.min(10, epic.countSwordType(arml, comb));
```
kei-gbf: sword had normalDamageLimit and ougiDamageLimit double?

src/global_logic.js
```diff
@@ -1692,8 +1694,31 @@ module.exports.addSkilldataToTotals = function (totals, comb, arml, buff) {
...
+                            totals[key]["cosmosNormalDamageLimit"] = Math.max(totals[key]["cosmosNormalDamageLimit"], 0.01 * Math.min(10, epic.countGunType(arml, comb)));
+                        } else if (amount == 'fist') {
+                            totals[key]["cosmosNormalDamageLimit"] = Math.max(totals[key]["cosmosNormalDamageLimit"], 0.01 * Math.min(10, epic.countFistType(arml, comb)));
+                        } else if (amount == 'bow') {
+                            totals[key]["cosmosNormalDamageLimit"] = Math.max(totals[key]["cosmosNormalDamageLimit"], 0.01 * Math.min(10, epic.countBowType(arml, comb)));
+                        } else if (amount == 'music') {
+                            totals[key]["cosmosNormalDamageLimit"] = Math.max(totals[key]["cosmosNormalDamageLimit"], 0.01 * Math.min(10, epic.countMusicType(arml, comb)));
+                        } else if (amount == 'katana') {
+                            totals[key]["cosmosNormalDamageLimit"] = Math.max(totals[key]["cosmosNormalDamageLimit"], 0.01 * Math.min(10, epic.countKatanaType(arml, comb)));
+                        }
```
kei-gbf: I suggest generic solution
- [ ] export genCountWeaponFunc

```js
const countWeaponType = epic.genCountWeaponFunc((arm) => arm.armType === amount);
totals[key]["cosmosNormalDamageLimit"] = Math.max(totals[key]["cosmosNormalDamageLimit"], 0.01 * Math.min(10, countWeaponType(arml, comb)));
```

This will reduce test and maintenance cost 10x
testers had to test those all 10 types weapon to ensure those code does not have mistake on copy and paste.
epic.js has unit test, unit test can guarantee if any type had pass the test then other types also ok.

- [ ] no need other countXXXTypes functions
if we wants add those functions, then need to add unit test to epic.test.js

src/global_logic.js
Groovman: nah just outdated code part from before I moved to cosmos damage limit up to check for the biggest cap up.

src/global_logic.js
Groovman: I rather have 10 functions from the same place near each other used in 100 different places as 100 times the same code with magic strings,,,

kagekiyo7:
otherChara is called Shield of Eternal Splendor category(総べ称号枠). 
I taked the liberty of changing the variable name :)
https://gbf.wiki/Wonders

Groovman:
You can add the wonder if you want to :D

src/global_logic.js
kei-gbf: that make extra maintenance and test cost to have a lot if-else branch and copy-pasted code
when we found bug then maintainer have to edit 10x places and tester check those all cases.

kei-gbf: about Math.max check, I don't know how it should be
are there verification source if two or more Cosmos weapon was set

current code behavior is:
- two cosmos sword -> no duplicate the effects
- cosmos sword and cosmos katana -> one of max count is used

Groovman: https://twitter.com/Hecate_mk2/status/1207917620080459776

kei-gbf: the behavior is ok
about the many copy-paste, you can think why it shouldn't 
it requires tests tasks each weapon types to check those behaviors.

kagekiyo7:
I updated Meteon data based on gamewith.
https://xn--bck3aza1a2if6kra4ee0hf.gamewith.jp/article/show/178713

Groovman:
> I updated Meteon data based on gamewith.
> https://xn--bck3aza1a2if6kra4ee0hf.gamewith.jp/article/show/178713
> 
> Gain absolute TA if off 2nd support.

Meteon support skill changes are looking wrong to me.
https://gbf.wiki/Meteon_(Holiday)

kagekiyo7:
Wiki is probably copy error.
image

Groovman:
> Wiki is probably copying mistake.
> image

110% boost to C.A. damage is not missing?

kagekiyo7:
The wiki references gamewith, and it isn't writing about "Enhanced mystery magnification to 10.5x".
I think the wiki editor has misunderstood "Enhanced mystery magnification to 10.5x" for "110% boost to C.A. damage" for some reason.

Groovman:
> The wiki references gamewith, and it isn't writing about "Enhanced mystery magnification to 10.5x".
> I think the wiki editor has misunderstood "Enhanced mystery magnification to 10.5x" for "110% boost to C.A. damage" for some reason.

En wiki displays the ougi ratio as base ratio and has the 110% to the support skill, Gamewith just display the resulting 1050% in the ougi ratio.

kagekiyo7:
To be precise,
950%(4.5x + 500%) x 110%(support) = 10.45...?

Should I change the ougi ratio to 9.5x and add support ougi 10%?

Groovman:
> To be precise,
> 950%(4.5x + 500%) x 110%(support) = 10.45...?
> 
> Should I change the ougi ratio to 9.5x and add support ougi 10%?

it is 500% base * 2.1 multiplier => 1050% resulting ougi ratio

kagekiyo7:
ok, my mistake.
Support ~110%~ 210%

kagekiyo7:
Hello, This PR has bug. #400 
Some personal buffs are not applied to the 2nd or later grids.
image

Savedata 
[motocal_data.txt](https://github.com/MotocalDevelopers/motocal/files/4008230/motocal_data.txt)

Groovman:
Latest changes should fix #400 , djeeta only fields were reused for support abilities, so resetting or not resetting them gave both wrong values either for MC or the character. I removed the whole reseting and just recreate a new totals map every time and read the buff out for each.

Also added some fixes changes from my latest dev build that were requested.

kagekiyo7:
x1 test: 6%?
image

PR #406, opened 2020-02-01
Patch 2020 01 17

kei-gbf:
New Weapons
  * [x] [Soshu Masamune](https://gbf.wiki/Soshu_Masamune)
  * [x] [Spectral Cleaver](https://gbf.wiki/Spectral_Cleaver)

New Characters
  * [x] [Jin (SSR)](https://gbf.wiki/Jin)
  * [x] [Danua (SSR)](https://gbf.wiki/Danua_(SSR))


Groovman:
I merged the recent changes and tested the update. Different to D.Jeanne and Sara, Danua and Black Knight have emnity in steps. I added that step system too, and updated the support skill for both. Text and Translation is the same for both skills. I need to update the translation after that it is ready for merging.

PR #411, opened 2020-03-09
Fix real hp export for NVD3

Groovman:
In real hp export some hp values are null, ignored them on export to NVD3.


Groovman:
Please merge, minimal change.

kei-gbf:
I will try but I had not touch NVD3
Can you show the incorrect case ? or how to re-produce the bug.

- [x] I just know the general case, this patch will avoid the null value
  - [x] checked graphData[i][j] != null
  - [ ] but no check graphData[i][0] ? (I dont have info what is the graphData)
- [ ] the patch will change length of array, I don't know it is ok or no?

Groovman:
> I will try but I had not touch NVD3
> Can you show the incorrect case ? or how to re-produce the bug.
> 
> * [x]  I just know the general case, this patch will avoid the null value
>   
>   * [x]  checked graphData[i][j] != null
>   * [ ]  but no check graphData[i][0] ? (I dont have info what is the graphData)
> * [ ]  the patch will change length of array, I don't know it is ok or no?

Just add 2 sets with different max hp, when you set it to real hp, one has more or less hp and because of that different x and y coordinates getting created. The x coords for all sets are in the same array. y coords that have not a related x coord are null and tull when exporting the array, x coords that have a y coord that is null have to be ignored when exporting on real hp.

PR #412, opened 2020-03-15
Fix Sunya having wrong elemental ATK Up

Groovman:
Oversight Sunya should have Elemental ATK large but had med.

Skill name was listed under med skill, large skill was not added.

Fixed now.


Groovman:
Will be included in the patch for 2020-03-17, because needed for Mugen's harp.

PR #427, opened 2020-06-15
Updates till Patch 2020-06-13

Groovman:
https://gbf.wiki/News_2020
2020-03-20 -> 2020-06-13


OrkunKocyigit:
From what I see, 4* (Ancient) Perseus maxHP numbers are incorrect.

PR #434, opened 2020-08-09
Add missing summedAttack processing

kagekiyo7:
> ①　攻撃力表記値/10、切り上げ
　65981→6599
②　①*艇、切り上げ
　6599*1.1=7258.9→7259
③　②*炉、切り上げ
　7259*1.1=7984.9→7985
④　③*10
　7985→79850

from: https://twitter.com/zekasyuz/status/1287157981000343552

I create new displayAttack variable for display.
It will branched to summedAttack variable.

imageimage

and I change to Math.ceil to Math.round

## Verification of "Math.round"
Sum of other numbers: 20715
Number of weapon: 1
Master Bonus ATK: 18%

- (20715 + (2950 x 1.2)) x 1.18 = 28620.9
image

- (20715 + (3300 x 1.2)) x 1.18 = 29116.5
image

- (20715 + (2480 x 1.2)) x 1.18 = 27955.38
image

- (20715 + (840 x 1.2)) x 1.18 = 25633.14
image

### DIFF
--- a/src/global_logic.js
+++ b/src/global_logic.js
@@ -1033,7 +1033,7 @@ module.exports.calcBasedOneSummon = function (summonind, prof, buff, totals) {
 
         res[key] = {
             totalAttack: Math.ceil(totalAttack),
-            displayAttack: Math.ceil(displayAttack),
+            displayAttack: Math.round(displayAttack),
             totalSkillCoeff: totalSkillCoeff,
             totalHP: Math.round(totalHP),
             displayHP: Math.round(displayHP),
--- a/src/global_logic.js
+++ b/src/global_logic.js
@@ -478,11 +478,11 @@ module.exports.calcBasedOneSummon = function (summonind, prof, buff, totals) {
         if (key == "Djeeta") {
             // for Djeeta
             // ATK
-            var summedAttack = totals[key]["baseAttack"];
-            summedAttack += totals[key]["armAttack"];
-            summedAttack += totalSummon["attack"];
-            summedAttack += totals["Djeeta"]["job"].atBonus;
-            summedAttack *= 1.0 + buff["master"];
+            var displayAttack = totals[key]["baseAttack"];
+            displayAttack += totals[key]["armAttack"];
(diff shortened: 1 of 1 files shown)


OrkunKocyigit:
It might be unrelated but in code line 3312 uses following statement:
`var summedAttack = onedata[key].displayAttack`
And the one that sets up this value is at line 1036
`displayAttack: Math.round(displayAttack),`
However line 508 defines summed attack as follows:
```
let summedAttack = Math.ceil(displayAttack / 10);
// TODO: Add Airship Effect and Crew Skill here.
//summedAttack = Math.ceil(summedAttack * (1.0 + Airship Effect));
//summedAttack = Math.ceil(summedAttack * (1.0 + Crew Skill));
summedAttack *= 10;
```
Question is do we need to update line 3312 as well?

kagekiyo7:
Yes. Thank you for reporting.

OrkunKocyigit:
I feel like calculations need to be done on display account variable then summed attack variable right? Let's say we get attack of 4998.5. In that case we set display attack to 4998.5, summed attack to 4999. Then calculate everything using 4999. Then pass 4998.5 to result.js as a display attack. Unless I am mistaken, shouldn't it be other way around?

kagekiyo7:
4998.5 of attack, becomes 4999 by `Math.round()` in case of `displayAttack`, (and passed to results.js)
https://github.com/MotocalDevelopers/motocal/blob/19234919a54f3c2ce07cc673ffb0266a35d13c16/src/global_logic.js#L1036

next, becomes 5000 by this code in case of `summedAttack`. (and used for calculation)
https://github.com/MotocalDevelopers/motocal/blob/19234919a54f3c2ce07cc673ffb0266a35d13c16/src/global_logic.js#L508-L512

Math.ceil(4998.5 / 10) = 500
500 * 10 = 5000

Do you feel a problem it?

OrkunKocyigit:
I checked on it with Twitter and ingame. It looks fine.

PR #440, opened 2020-09-02
Updates until 31-08-20

OrkunKocyigit:


Groovman (approved):
LGTM!

Because of renaming by Cygames some emnity skills were parsed wrong and stamina large was missing.

PR #441, opened 2020-09-05
Fix exHP-S skillSize

kagekiyo7:
From gamewith
https://xn--bck3aza1a2if6kra4ee0hf.gamewith.jp/article/show/71962

image
image


Groovman:
The values are right, it is probably cygames changed how HP is displayed, big HP is medium in some cases now and medium is small and small is just another small. So Xenos in Motocal are probably just exATK/HPM and not exATK/HPS.

Groovman:
Yeah exHP the last 5 need to be updated, S is fine, And "strengthLLandHPS" skill needs the amount changed to "LL-M".

kagekiyo7:
I have now verified exHP-M size.

exHP-M SLv1 size = 6.0%
　( Math.ceil( 7094 * 1.06 ) = 7520 )
exHP-M SLv2 size = 7.0%
　( Math.ceil( 7122 * 1.07 ) = 7621 )


Then you might guess that the list would look like this. I don't know exHP-L because i don't have the weapon. (and no verification data in internet)

        "S": [3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0, 10.0, 11.0, 12.0, 12.5, 13.0, 13.5, 14.0, 14.5, 14.8, 15.1, 15.4, 15.7, 16.0],
        "M":  [6.0, 7.0, 8.0, 9.0, 10.0, 11.0, 12.0, 13.0, 14.0, 15.0, 15.6, 16.2, 16.8, 17.4, 18.0, 18.4, 18.8, 19.2, 19.6, 20.0],
        "L":  ???,


## Verification process
### exHP-M SLv1.
image
image
image
image

### exHP-M SLv2.
image
image

Groovman:
https://gbf.wiki/Silverslime_Bell if you need an exHPL weapon

The thing I assume exHP-S is the mysterious one here that can be boosted by mysterious summons https://gbf.wiki/Mysterious_Hatchet
exATK+exHPS on xeno is most likely exATK+exHPM (in motocal). Cygames is using the same wording for different ex skills basically so the mods are more in line with the wording for normalHP.
So yeah like you found exHPM starts at 6 and https://gbf.wiki/Weapon_Skills#HP_Up says sl10 at 12 and 14.5 at sl15 and so on. The problem is with how cygames did changes to naming. basically the backend const vars need to be updated to handle the new naming correctly so we don't have to think everytime if the small means small or medium, or we adapt every time. We had the issue already with magna weapons where magnaHPS means actually magnaHPM.

kagekiyo7:
Sorry, I deleted the post once because I used the wrong normal-HP-M weapon.
I don't have a EX-HP skill weapon so I can't verify it.

----

OK. I will close this PR once.
and I change to exATK/HPM.
The problem of skill notation is very complex.

Groovman:
No need to apologize, I had the problem when reviewing the weapons magna 2, too. We need to change it at some point, because it will become really confusing in the future. It good that you found the issue with the Xeno weapons also.

PR #442, opened 2020-09-05
Fix EX-HP-M skill size / change strengthLLandHPS amount

kagekiyo7:
From gamewith
https://xn--bck3aza1a2if6kra4ee0hf.gamewith.jp/article/show/71962

image
image


Groovman (approved):
LGTM!

PR #465, opened 2021-09-09
Add LB Konshin/Haisui (Stamina/Emnity) support

shinraminagi:
- Got LB Konshin/Haisui feature, finally!
- Currently, 2 Konshins and 2 Haisuis can be set at maximum


kagekiyo7 (approved):
LGTM!

PR #466, opened 2021-09-10
Add Awakening Types and Awakening Lv8

shinraminagi:
- Support all 4 Awakening Types: "Balanced", "Attack", "Defense", and "Multiattack"
- Now, you can choose Awakening Lv up to 8!


kagekiyo7 (approved):
LGTM!

PR #469, opened 2021-11-16
Patch for Node v17

shinraminagi:
Node v17 fails to run webpack-dev-server with error:0308010C.
Set NODE_OPTIONS env var as a tentative workaround.


kagekiyo7 (approved):
LGTM!

Groovman:
I would put this on hold till node 17 changes are in LTS.

